PR #6, opened 2018-08-01
Parse known Gutenberg blocks and extract strings for translation

brucepearson:
Parse known Gutenberg blocks and extract strings for translation - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5653


brucepearson:
@OnTheGoSystems/wpml Please review

src/class-wpml-gutenberg-strings-in-block.php
```diff
@@ -0,0 +1,213 @@
+<?php
+
+/**
+ * Class WPML_Gutenberg_Strings_In_Block
+ */
+class WPML_Gutenberg_Strings_In_Block {
+
+	/** @var array */
+	private $block_types = array(
```
brucepearson: **Note:** These are xpath queries to find the element or attribute that has text needing translation. These will come from `wpml-config.xml` and will be handled in the following merge request.

src/class-wpml-gutenberg-strings-in-block.php
```diff
@@ -0,0 +1,213 @@
...
+		'core/paragraph' => array( '//p' ),
+		'core/heading'   => array( "//*[name()='h1' or name()='h2' or name()='h3' or name()='h4' or name()='h5' or name()='h6']" ),
```
andreasciamanna: @BruceOnTheGo I think that's a better way to do the same: `//*[self:: h1 or self:: h2 or self::h3 or self::h4 or self::h5 or self::h6]`.

It also looks like your solution may have unpredictable effects in some circumstances: https://stackoverflow.com/a/722251/626538

andreasciamanna (approved):
It looks ok. I'm not 100% sure about the xPath query used for `h*` elements, though.

src/class-wpml-gutenberg-strings-in-block.php
brucepearson: @andreasciamanna I've updated the query in the wpml-config.xml file - https://git.onthegosystems.com/wpml/sitepress-multilingual-cms/merge_requests/1986/diffs

PR #8, opened 2018-08-07
Fix rendering of Gutenberg blocks

brucepearson:
Fix rendering of Gutenberg blocks - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5679

This is required because the Gutenberg function `gutenberg_render_block` doesn't render the block structure `<-- wp:xxxx ..... -->` that is required by the editor.

Gutenberg doesn't have a function to do this. For now we'll have to render it ourselves. Then I'll see if we can supply a fix for Gutenberg so they can add it to their code.

### DIFF
--- a/src/class-wpml-gutenberg-integration.php
+++ b/src/class-wpml-gutenberg-integration.php
@@ -28,7 +28,7 @@ class WPML_Gutenberg_Integration {
 		WPML_Gutenberg_Config_Option $config_option
 	) {
 		$this->strings_in_blocks = $strings_in_block;
-		$this->config_option= $config_option;
+		$this->config_option     = $config_option;
 	}
 
 	public function add_hooks() {
@@ -120,7 +120,7 @@ class WPML_Gutenberg_Integration {
 
 			$content = '';
 			foreach ( $blocks as $block ) {
-				$content .= gutenberg_render_block( $block );
+				$content .= $this->render_block( $block );
 			}
 
 			wp_update_post( array( 'ID' => $translated_post_id, 'post_content' => $content ) );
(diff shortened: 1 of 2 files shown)


src/class-wpml-gutenberg-integration.php
```diff
@@ -152,6 +155,71 @@ private function update_block_translations( $blocks, $string_translations, $lang
...
+		if ( isset( $block['blockName'] ) ) {
+			$block_type = substr( $block['blockName'], 5 ); // strip core. eg core/paragraph
+
+			$block_attributes = '';
+			if ( $block['attrs'] ) {
+				$block_attributes = ' ' . json_encode( $block['attrs'] );
+			}
+			$content .= '<!-- wp:' . $block_type . $block_attributes . ' -->';
+
+			if ( isset ( $block['innerBlocks'] ) ) {
```
brucepearson: *Note:* This part will be changed in the next task because it doesn't handle the Gutenberg columns block fully.

src/class-wpml-gutenberg-integration.php
```diff
@@ -152,6 +155,71 @@ private function update_block_translations( $blocks, $string_translations, $lang
...
+	/**
+	 * @param array $block
+	 *
+	 * @return string
+	 */
+	private function render_block( $block ) {
+		$content = '';
+
+		if ( isset( $block['blockName'] ) ) {
+			$block_type = substr( $block['blockName'], 5 ); // strip core. eg core/paragraph
```
strategio: @brucepearson, will this method render only block from "core"? In that case, why not `preg_replace( '/^core\//', '', $block['blockName'] )` and remove the comment?

brucepearson: @strategio Good idea. I've changed this.
We're only handling `core` now because that seems to be the only type Gutenberg is handling in its parser from what I can see.

strategio (approved):

PR #10, opened 2018-08-13
Don't include gutenberg columns and column inner HTML in translation editor

brucepearson:
Don't include gutenberg columns and column inner HTML in translation editor - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5685


gordonsh24 (approved):

strategio (approved):
+1 @brucepearson

PR #12, opened 2018-09-13
Fix fatal error with Gutenberg 3.8 - https://onthegosystems.myjetbrai…

brucepearson:
Fix fatal error with Gutenberg 3.8 - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5825

### DIFF
--- a/src/class-wpml-gutenberg-integration.php
+++ b/src/class-wpml-gutenberg-integration.php
@@ -73,28 +73,30 @@ class WPML_Gutenberg_Integration {
 	 * @param array $blocks
 	 * @param array $package_data
 	 */
-	private function register_blocks( $blocks, $package_data ) {
+	private function register_blocks( array $blocks, array $package_data ) {
 
 		foreach ( $blocks as $block ) {
-			$strings = $this->strings_in_blocks->find( $block );
 
-			foreach ( $strings as $string ) {
+			if ( $block instanceof WP_Block_Parser_Block ) {
+				$strings = $this->strings_in_blocks->find( $block );
 
-				do_action(
-					'wpml_register_string',
-					$string->value,
-					$string->id,
-					$package_data,
-					$string->name,
-					$string->type
-				);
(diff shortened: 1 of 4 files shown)


src/class-wpml-gutenberg-integration.php
```diff
@@ -76,25 +76,27 @@ function register_strings( WP_Post $post, $package_data ) {
 	private function register_blocks( $blocks, $package_data ) {
```
andreasciamanna: @BruceOnTheGo can you add `array` type-hint to both arguments, please?

andreasciamanna: In the PHPDoc, it would help to be a bit more specific and define `@param WP_Block_Parser_Block[] $blocks`

andreasciamanna (dismissed):
+1

src/class-wpml-gutenberg-integration.php
brucepearson: I added `array` type hints. 

$blocks is an array of mixed items and only some of them are `WP_Block_Parser_Block`

andreasciamanna (dismissed):

src/class-wpml-gutenberg-integration.php
```diff
@@ -138,37 +140,39 @@ public function string_translated(
...
+				}
 			}
 		}
 
 		return $blocks;
 	}
 
 	/**
-	 * @param array $block
+	 * @param array|WP_Block_Parser_Block $block
```
strategio, line 163: Can it still be an array @brucepearson?

strategio (dismissed):
Looks good @brucepearson +1

src/class-wpml-gutenberg-integration.php
brucepearson: @strategio Yes, it can be an array. Gutenberg returns an array with `innerHTML` for anything that's not inside a block.

PR #13, opened 2018-09-13
Fix problem with inner image in paragraph block

brucepearson:
Fix problem with inner image in paragraph block - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmltm-2766

### DIFF
--- a/src/class-wpml-gutenberg-strings-in-block.php
+++ b/src/class-wpml-gutenberg-strings-in-block.php
@@ -74,7 +74,7 @@ class WPML_Gutenberg_Strings_In_Block {
 
 		if ( $block_queries ) {
 
-			$dom   = $this->get_dom( $block );
+			$dom   = $this->get_dom( $block->innerHTML );
 			$xpath = new DOMXPath( $dom );
 			foreach ( $block_queries as $query ) {
 				$elements = $xpath->query( $query );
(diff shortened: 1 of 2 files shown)


src/class-wpml-gutenberg-strings-in-block.php
```diff
@@ -162,10 +162,13 @@ private function set_element_value( DOMNode $element, $value ) {
 		if ( $element instanceof DOMAttr ) {
 			$element->parentNode->setAttribute( $element->name, $value );
 		} else {
-			$fragment = $element->ownerDocument->createDocumentFragment();
-			$fragment->appendXML( $value );
-			$clone = $element->cloneNode(); // Get element copy without children
-			$clone->appendChild( $fragment );
+			$clone = $element->cloneNode( false ); // Get element copy without children
```
andreasciamanna: @BruceOnTheGo I either need to read what `false` means or rely in your comment.

Wouldn't be better to create a `cloneNodeWithoutChildren()` method which works as a wrapper of `$element->cloneNode( false )`?

That way you don't need to add any comment here.

andreasciamanna (approved):

src/class-wpml-gutenberg-strings-in-block.php
brucepearson: Done

strategio (approved):
+1 @brucepearson

PR #16, opened 2018-10-25
Fix for WP 5.0

brucepearson:
Fix for WP 5.0 - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-5959

### DIFF
--- a/src/class-wpml-gutenberg-integration.php
+++ b/src/class-wpml-gutenberg-integration.php
@@ -63,7 +63,7 @@ class WPML_Gutenberg_Integration {
 			do_action( 'wpml_start_string_package_registration', $package_data );
 
 			$this->register_blocks(
-				gutenberg_parse_blocks( $post->post_content ),
+				$this->parse_blocks( $post->post_content ),
 				$package_data
 			);
 
@@ -124,7 +124,7 @@ class WPML_Gutenberg_Integration {
 	) {
 
 		if ( self::PACKAGE_ID === $package_kind ) {
-			$blocks = gutenberg_parse_blocks( $original_post->post_content );
+			$blocks = $this->parse_blocks( $original_post->post_content );
 
 			$blocks = $this->update_block_translations( $blocks, $string_translations, $lang );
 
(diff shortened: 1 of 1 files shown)


kagg-design (approved):
+1

src/class-wpml-gutenberg-integration.php
```diff
@@ -298,4 +298,15 @@ public function should_body_be_translated_filter( $translate, WP_Post $post, $co
 	private function is_gutenberg_post( WP_Post $post ) {
 		return (bool) preg_match( '/' . self::GUTENBERG_OPENING_START . '/', $post->post_content );
 	}
+
+	private function parse_blocks( $content ) {
+		global $wp_version;
+		if ( version_compare( $wp_version, '5.0-beta1', '>=' ) ) {
```
andreasciamanna, line 304: @BruceOnTheGo why you can't use `version_compare( $wp_version, '5.0', '>=' )`?

andreasciamanna (approved):
It's ok, but I'm not convinced about checking against 5.0-beta1

src/class-wpml-gutenberg-integration.php
brucepearson: @andreasciamanna Because beta versions are considered to be less than version without beta.

eg.
`version_compare( '5.0-beta1', '5.0', '>=' );` will return `false`

http://php.net/manual/en/function.version-compare.php

andreasciamanna: Right. Now I remember. 

So we should change this once 5.0 becomes official.

kagg-design: Yes, I also checked this `version_compare( '5.0-beta1', '5.0', '>=' )` when put +1

dimitrismitsis: For the record, I also checked this and I don't think it needs to be adjusted before the 5.0 release.  
This also works properly when used with the Gutenberg plugin (WordPress 4.9.8 currently)

PR #19, opened 2018-11-19
Decode HTML entities in the strings to translate

strategio:
This applies only for LINE and AREA because VISUAL fields will be
automatically decoded by TinyMCE.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6066

### DIFF
--- a/tests/phpunit/tests/test-wpml-gutenberg-stings-in-block.php
+++ b/tests/phpunit/tests/test-wpml-gutenberg-stings-in-block.php
@@ -10,6 +10,7 @@ class Test_WPML_Gutenberg_Strings_In_Block extends OTGS_TestCase {
 
 	/**
 	 * @test
+	 * @group wpmlcore-6066
 	 */
 	public function it_finds_paragraph() {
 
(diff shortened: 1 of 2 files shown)


tests/phpunit/tests/test-wpml-gutenberg-stings-in-block.php
```diff
@@ -19,7 +20,7 @@ public function it_finds_paragraph() {
 
 		$strings_in_block = new WPML_Gutenberg_Strings_In_Block( $config_option );
 
-		$paragraph = 'some paragraph';
+		$paragraph = 'some paragraph &amp special chars';
```
brucepearson: @strategio Should this be `&amp` or `&amp;`?

brucepearson (dismissed):
@strategio This is ok. +1 Just one question about the test.

adrianosferreira (dismissed):
+1

tests/phpunit/tests/test-wpml-gutenberg-stings-in-block.php
strategio: Good catch @brucepearson :)

This is corrected.

adrianosferreira (dismissed):
+1

PR #20, opened 2018-11-20
Convert the emojis into HTML entities before to load it into the `DOMDocument`

strategio:
Took inspiration from https://stackoverflow.com/a/8218649/2377680.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6070

### DIFF
--- a/tests/phpunit/tests/test-wpml-gutenberg-stings-in-block.php
+++ b/tests/phpunit/tests/test-wpml-gutenberg-stings-in-block.php
@@ -20,7 +20,7 @@ class Test_WPML_Gutenberg_Strings_In_Block extends OTGS_TestCase {
 
 		$strings_in_block = new WPML_Gutenberg_Strings_In_Block( $config_option );
 
-		$paragraph = 'some paragraph &amp; special chars';
+		$paragraph = 'some paragraph &amp; special chars &amp; emoji ❤ 😀 👍️';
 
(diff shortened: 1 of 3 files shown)


tests/phpunit/tests/test-wpml-gutenberg-stings-in-block.php
```diff
@@ -20,7 +20,7 @@ public function it_finds_paragraph() {
 
 		$strings_in_block = new WPML_Gutenberg_Strings_In_Block( $config_option );
 
-		$paragraph = 'some paragraph &amp; special chars';
+		$paragraph = 'some paragraph &amp; special chars &amp; emoji ❤';
```
andreasciamanna: @strategio is it GitHub or this is not an emoji?

I see a simple [hearth symbol](
https://www.rapidtables.com/code/text/ascii/ascii-heart.html) (♥) instead of something like 💛

brucepearson (dismissed):
@strategio This looks ok. +1

andreasciamanna (dismissed):

andreasciamanna:
I've pushed a minor change because, as explained in the commit, the problem was reported with emojis, therefore, it makes sense to use actual emojis in tests.

dimitrismitsis:
@andreasciamanna did you close this PR purposefully?

PR #21, opened 2018-11-20
Support block parsing for classic blocks

strategio:
Classic blocks are not wrapped with the usual `<!-- wp:block -->`
comments. But it's parsed by WP (in `parse_blocks`) which returns a block
without name.

In order to also process those classic blocks in the translation editor,
we will temporarily give it a name so it can be handled as any other
block.

block editor

classic block

At rendering time, we will process it as a raw HTML and apply `wpautop`
as it depends on if it was translated with the visual tab or text tab (note that it can only be edited with the visual in the block editor).

Unfortunately, I could not add test because the `WP_Block_Parser_Block`
class is hardcoded inside our code.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6075

### DIFF
--- a/src/class-wpml-gutenberg-integration.php
+++ b/src/class-wpml-gutenberg-integration.php
@@ -8,6 +8,7 @@ class WPML_Gutenberg_Integration {
 	const PACKAGE_ID              = 'Gutenberg';
 	const GUTENBERG_OPENING_START = '<!-- wp:';
 	const GUTENBERG_CLOSING_START = '<!-- /wp:';
+	const CLASSIC_BLOCK_NAME      = 'core/classic-block';
 
 	/**
 	 * @var WPML_Gutenberg_Strings_In_Block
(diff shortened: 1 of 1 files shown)


src/class-wpml-gutenberg-integration.php
```diff
@@ -80,32 +81,44 @@ private function register_blocks( array $blocks, array $package_data ) {
...
+		}
+	}
 
-				}
+	/**
+	 * @param WP_Block_Parser_Block|array $block
+	 *
+	 * @return WP_Block_Parser_Block
+	 */
+	private function sanitize_block( $block ) {
```
strategio, line 111: Passing through this method will ensure that we get a `WP_Block_Parser_Block` object and we can process it as such.

brucepearson (approved):
@strategio I did some testing and this looks ok. +1

PR #22, opened 2018-11-21
Added special case for guessing inner HTML parts for block core/media-text

adrianosferreira:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6074

### DIFF
--- a/src/class-wpml-gutenberg-integration.php
+++ b/src/class-wpml-gutenberg-integration.php
@@ -257,17 +257,30 @@ class WPML_Gutenberg_Integration {
 
 		$parts = array( $inner_HTML, '' );
 
-		preg_match( '#>\s*</#', $inner_HTML, $matches );
+		switch( $block->blockName ) {
+			case 'core/media-text':
+				$html_to_find = '<div class="wp-block-media-text__content">';
+				$pos          = mb_strpos( $inner_HTML, $html_to_find ) + mb_strlen( $html_to_find );
+				$parts        = array(
+					mb_substr( $inner_HTML, 0, $pos ),
+					mb_substr( $inner_HTML, $pos )
+				);
+				break;
 
-		if ( count( $matches ) === 1 ) {
-			$parts = explode( $matches[0], $inner_HTML );
-			if ( count( $parts ) === 2 ) {
-				$match_mid_point = 1 + ( strlen( $matches[0] ) - 3 ) / 2;
-				// This is the first ">" char plus half the remaining between the tags
+			default:
+				preg_match( '#>\s*</#', $inner_HTML, $matches );
 
-				$parts[0]        .= substr( $matches[0], 0, $match_mid_point );
-				$parts[1]        = substr( $matches[0], $match_mid_point ) . $parts[1];
-			}
+				if ( count( $matches ) === 1 ) {
+					$parts = explode( $matches[0], $inner_HTML );
+					if ( count( $parts ) === 2 ) {
+						$match_mid_point = 1 + ( mb_strlen( $matches[0] ) - 3 ) / 2;
+						// This is the first ">" char plus half the remaining between the tags
+
+						$parts[0] .= mb_substr( $matches[0], 0, $match_mid_point );
+						$parts[1] = mb_substr( $matches[0], $match_mid_point ) . $parts[1];
+					}
+				}
+				break;
 		}
 
 		return $parts;
--- a/tests/phpunit/tests/test-wpml-gutenberg-integration.php
+++ b/tests/phpunit/tests/test-wpml-gutenberg-integration.php
@@ -336,6 +336,84 @@ class Test_WPML_Gutenberg_Integration extends OTGS_TestCase {
 
 	}
 
+	/**
+	 * @test
+	 * @group wpmlcore-6074
+	 */
+	public function it_updates_media_text_inner_blocks() {
+		$inner_html = '<div class="wp-block-media-text__content"></div>';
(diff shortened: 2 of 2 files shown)


adrianosferreira:

src/class-wpml-gutenberg-integration.php
```diff
@@ -257,17 +257,30 @@ private function guess_inner_HTML_parts( $block ) {
...
-				$match_mid_point = 1 + ( strlen( $matches[0] ) - 3 ) / 2;
-				// This is the first ">" char plus half the remaining between the tags
-
-				$parts[0]        .= substr( $matches[0], 0, $match_mid_point );
-				$parts[1]        = substr( $matches[0], $match_mid_point ) . $parts[1];
-			}
+		switch( $block->blockName ) {
+			case 'core/media-text':
+				$html_to_find = '<div class="wp-block-media-text__content">';
+				$pos          = strpos( $inner_HTML, $html_to_find ) + strlen( $html_to_find );
```
strategio: @adrianosferreira, you should use multi-bytes functions `mb_strpos`, `mb_strlen` and `mb_substr`.

We also need to fix this in the `default` logic.

You can add some exotic chars in the tests.

strategio (changes requested):

src/class-wpml-gutenberg-integration.php
adrianosferreira: @strategio I don't see the need for that. I can see non-multibyte functions all over the place in the Gutenberg library and 0 occurrences of multibyte ones, so I guess we should change the whole lib. I guess that @BruceOnTheGo should have a reason for that.

Anyway, this code is just a suggestion from Bruce, I believe he had a reason to write using normal string functions.

strategio: @adrianosferreira, I don't see disadvantage using multi-byte functions (some will complain about performance but it's not significant IMO), it's just safer and we already had a lot of bugs related to this.

If changing the functions to multi-byte in the code base does not break the existing tests, I would go for it.

src/class-wpml-gutenberg-integration.php
```diff
@@ -257,17 +257,30 @@ private function guess_inner_HTML_parts( $block ) {
...
-			$parts = explode( $matches[0], $inner_HTML );
-			if ( count( $parts ) === 2 ) {
-				$match_mid_point = 1 + ( strlen( $matches[0] ) - 3 ) / 2;
-				// This is the first ">" char plus half the remaining between the tags
-
-				$parts[0]        .= substr( $matches[0], 0, $match_mid_point );
-				$parts[1]        = substr( $matches[0], $match_mid_point ) . $parts[1];
-			}
+		switch( $block->blockName ) {
+			case 'core/media-text':
```
strategio, line 261: @adrianosferreira, you wrote a test for this case but if you put a breakpoint on the line below (262), you will see that it never stops.

strategio (changes requested):

src/class-wpml-gutenberg-integration.php
adrianosferreira: @strategio if the multibyte thing is a blocking issue for you I will hold this fix until @BruceOnTheGo is online to check what he thinks.

I will check the test issue.

strategio: I'll leave the multi-bytes suggestion up to you @adrianosferreira (to not delay our release), not blocking.

src/class-wpml-gutenberg-integration.php
brucepearson: @strategio I keep forgetting to use the multi-byte functions. @adrianosferreira Please change to use the multi-byte functions. If the existing tests don't break we should be good to go.

adrianosferreira: @strategio adjusted both suggestions, please check.

src/class-wpml-gutenberg-integration.php
```diff
@@ -257,17 +257,30 @@ private function guess_inner_HTML_parts( $block ) {
...
+				$html_to_find = '<div class="wp-block-media-text__content">';
+				$pos          = mb_strpos( $inner_HTML, $html_to_find ) + mb_strlen( $html_to_find );
+				$parts        = array(
+					substr( $inner_HTML, 0, $pos ),
```
strategio: @adrianosferreira, you can use `mb_substr` here and below.

src/class-wpml-gutenberg-integration.php
```diff
@@ -257,17 +257,30 @@ private function guess_inner_HTML_parts( $block ) {
...
+				);
+				break;
+
+			default:
+				preg_match( '#>\s*</#', $inner_HTML, $matches );
+
+				if ( count( $matches ) === 1 ) {
+					$parts = explode( $matches[0], $inner_HTML );
+					if ( count( $parts ) === 2 ) {
+						$match_mid_point = 1 + ( strlen( $matches[0] ) - 3 ) / 2;
```
strategio: And here `mb_strlen`.

strategio (changes requested):

src/class-wpml-gutenberg-integration.php
adrianosferreira: done.

src/class-wpml-gutenberg-integration.php
adrianosferreira: done.

strategio (dismissed):
Sounds good @adrianosferreira +1

brucepearson (approved):

PR #23, opened 2018-11-21
Do not register strings from the block editor if the content has no block

strategio:
WP block parser wraps the raw HTML (with no meta-comment) into a block.
So we were creating a string with the whole post content in that case (side effect of wpmlcore-6075).

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6083


andreasciamanna (dismissed):

adrianosferreira (dismissed):
+1

PR #25, opened 2019-01-08
Force to update the post content in the target language

strategio:
We have a filter `WPML_Term_Adjust_Id::filter` which is messing the terms
when we are saving the post content. Sometimes it converts the
translated terms into the source ones (because the current lang is the
original one).

The filter is relying on several debug backtrace conditions which are
impacted depending on the activated page builders (e.g. we have a
problem when Enfold + Gutenberg are enabled, but not when only Gutenberg
is enabled).

By saving in the target lang, we ensure the terms are not messed even if
it's filtered.

Moreover, at this step we should not have nested post save so we should
not have side effects.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6221

### DIFF
--- a/src/class-wpml-gutenberg-integration-factory.php
+++ b/src/class-wpml-gutenberg-integration-factory.php
@@ -3,12 +3,15 @@
 class WPML_Gutenberg_Integration_Factory {
 
 	public function create() {
+		/** @var SitePress $sitepress */
+		global $sitepress;
 
 		$config_option = new WPML_Gutenberg_Config_Option();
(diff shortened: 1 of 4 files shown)


kagg-design (approved):
+1 @pierre

Igor-WPML (approved):
+1 @strategio

strategio:
One more review please @OnTheGoSystems/wpml.

gordonsh24 (approved):

PR #28, opened 2019-02-12
Improve handling of Inner blocks in Gutenberg

strategio:
We can now use a new property `WP_Block_Parser_Block::$innerContent`
which provides the sequence of inner elements: strings or
null if it's an inner block.

Then we just need to replace the null value with inner blocks.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6330

### DIFF
--- a/src/class-wpml-gutenberg-integration.php
+++ b/src/class-wpml-gutenberg-integration.php
@@ -197,15 +197,12 @@ class WPML_Gutenberg_Integration {
 	private function render_inner_HTML( $block ) {
 
 		if ( isset ( $block->innerBlocks ) && count( $block->innerBlocks ) ) {
-			$inner_html_parts = $this->guess_inner_HTML_parts( $block );
 
-			$content = $inner_html_parts[0];
-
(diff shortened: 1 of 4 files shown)


src/class-wpml-gutenberg-integration.php
```diff
@@ -247,6 +244,53 @@ private function render_inner_HTML( $block ) {
...
+			$content .= $this->render_block( $inner_block );
+		}
+
+		$content .= $inner_html_parts[1];
+
+		return $content;
+	}
+
 	/**
 	 * The gutenberg parser doesn't handle inner blocks correctly
```
brucepearson: @strategio Please update this comment to include the WP versions it's for: `The gutenberg parser prior to version x.xx doesn't......`

brucepearson (dismissed):
@strategio This is good. Nice and straight forward now! +1

strategio:
@OnTheGoSystems/wpml give it a look please!

kagg-design (dismissed):
+1 @strategio

andreasciamanna (dismissed):

PR #29, opened 2019-02-19
Mark headings in XLIFFs so that ATE can indicate that these fields have SEO significance

kagg-design:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmltm-3081

### DIFF
--- a/src/class-wpml-gutenberg-strings-registration.php
+++ b/src/class-wpml-gutenberg-strings-registration.php
@@ -77,6 +77,15 @@ class WPML_Gutenberg_Strings_Registration {
 				);
 
 				$this->update_string_location( $package_data, $string );
+
+				// phpcs:disable WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+				if ( 'core/heading' === $block->blockName ) {
+					// phpcs:enable
+					$wrap_tag = (string) isset( $block->attrs['level'] ) ? $block->attrs['level'] : 2;
+					$wrap_tag = 'h' . $wrap_tag;
+					$this->update_wrap_tag( $package_data, $string, $wrap_tag );
+				}
+
 				$this->remove_string_from_leftovers( $string->value );
 			}
 
@@ -90,9 +99,26 @@ class WPML_Gutenberg_Strings_Registration {
 		$string_id = apply_filters( 'wpml_string_id_from_package', 0, $package_data, $string_data->id, $string_data->value );
 		$string    = $this->string_factory->find_by_id( $string_id );
 
-		if ( $string_id && $string ) {
+		if ( $string_id ) {
 			$string->set_location( $this->string_location );
 			$this->string_location++;
+		}
+	}
+
+	/**
+	 * Update string wrap tag.
+	 * Used for SEO, can contain (h1...h6, etc.)
+	 *
+	 * @param array    $package_data Package.
+	 * @param stdClass $string_data  String in the package.
+	 * @param string   $wrap_tag     String wrap.
+	 */
+	private function update_wrap_tag( $package_data, stdClass $string_data, $wrap_tag ) {
+		$string_id = apply_filters( 'wpml_string_id_from_package', 0, $package_data, $string_data->id, $string_data->value );
+		$string    = $this->string_factory->find_by_id( $string_id );
+
+		if ( $string_id ) {
+			$string->set_wrap_tag( $wrap_tag );
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


src/class-wpml-gutenberg-strings-registration.php
```diff
@@ -96,6 +105,23 @@ private function update_string_location( array $package_data, stdClass $string_d
...
+	 * Update string wrap tag.
+	 * Used for SEO, can contain (h1...h6, etc.)
+	 *
+	 * @param array    $package_data Package.
+	 * @param stdClass $string_data  String in the package.
+	 * @param string   $wrap_tag     String wrap.
+	 */
+	private function update_wrap_tag( $package_data, stdClass $string_data, $wrap_tag ) {
+		$string_id = apply_filters( 'wpml_string_id_from_package', 0, $package_data, $string_data->id, $string_data->value );
+		$string    = $this->string_factory->find_by_id( $string_id );
```
brucepearson, line 118: @kagg-design Shouldn't you check the $string_id here instead of line 120?

brucepearson (dismissed):
@kagg-design This looks ok. +1

src/class-wpml-gutenberg-strings-registration.php
kagg-design: @brucepearson you are right, its overkill. find_by_id() always returns instance of WPML_ST_String::

Fixed it.

tests/phpunit/tests/test-wpml-gutenberg-string-registration.php
```diff
@@ -143,6 +143,64 @@ public function it_registers_strings_and_set_location() {
 		$subject->register_strings( $post, $package );
 	}
 
+	/**
+	 * @test
+	 * @group wpmltm-3081
+	 */
+	public function it_registers_strings_and_set_wrap_tag() {
+		$post               = \Mockery::mock( 'WP_Post' );
+		$post->post_content = '<!-- wp:something -->post content<!-- /wp:something -->';
```
andreasciamanna: @kagg-design I think this line is a bit misleading.

In theory, you can put whatever you want here, right? Because `gutenberg_parse_blocks` will overrides whatever you put in the post content.

If I'm right, wouldn't be clearer to have something like that instead?

`$post->post_content = 'post content is not relevant in this test';`

tests/phpunit/tests/test-wpml-gutenberg-string-registration.php
```diff
@@ -143,6 +143,64 @@ public function it_registers_strings_and_set_location() {
...
+		$blocks['block 1']->attrs['level'] = 1;
+		$string_name_1                     = md5( $blocks['block 1']->blockName . $blocks['block 1']->innerHTML );
+		$string_id_1                       = 123;
+
+		$blocks['block 2']                 = \Mockery::mock( 'WP_Block_Parser_Block' );
+		$blocks['block 2']->blockName      = 'core/heading';
+		$blocks['block 2']->innerHTML      = 'some block content 2';
+		$blocks['block 2']->attrs['level'] = 2;
+		$string_name_2                     = md5( $blocks['block 2']->blockName . $blocks['block 2']->innerHTML );
+		$string_id_2                       = 456;
```
andreasciamanna: @kagg-design I would reduce some duplication here.

1. Create a private helper function:  
   ```php
	/**
	 * @return \WP_Block_Parser_Block $block|\Mockery\MockInterface
	 */
	private function get_block( $name, $html ) {
		/** @var \WP_Block_Parser_Block $block|\Mockery\MockInterface */
		$block            = \Mockery::mock( 'WP_Block_Parser_Block' );
		$block->blockName = $name;
		$block->innerHTML = $html;


		return $block;
	}
   ```
2. Use it:  
   ```php
		$blocks = array(
			'block 1' => $this->get_block( 'some name 1', 'some block content 1' ),
			'block 2' => $this->get_block( 'some name 2', 'some block content 2' ),
		);

		$string_name_1 = md5( $blocks['block 1']->blockName . $blocks['block 1']->innerHTML );
		$string_id_1   = 123;

		$string_name_2 = md5( $blocks['block 2']->blockName . $blocks['block 2']->innerHTML );
		$string_id_2   = 456;
   ```

andreasciamanna (changes requested):

tests/phpunit/tests/test-wpml-gutenberg-string-registration.php
kagg-design: @andreasciamanna sure, done in all test functions.

tests/phpunit/tests/test-wpml-gutenberg-string-registration.php
kagg-design: @andreasciamanna thank you for pointing on it. Done in my test and all already existing tests. Commit 18847698533c5374042f319e53aac068678b1aeb

PR #32, opened 2019-04-25
Include the reusable string packages in the translation job of the document it's inserted in.

strategio:
When the reusable block is saved, strings will be created in a package
attached to the reusable block ID. We will add this package to the main
package of strings attached to the post we are translating.

In order to have a better organization of the code and decoupling, I
created the `Integration_Composite` class, so we can load several
integration classes.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6563

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,8 @@
   },
   "require": {
     "roave/security-advisories": "dev-master",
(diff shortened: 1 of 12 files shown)


strategio:
I am closing this PR because we will implement a different workflow.

PR #33, opened 2019-04-26
Create a translation post for each block included in the job

strategio:
Our existing code will handle the block packages as the main package,
but we just need to make sure that a post translation exist for each
block.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6564

PR #34, opened 2019-04-29
Convert the reusable block ID on the fly when it's rendered

strategio:
I decided to remove the code to convert the ID inside the translated
post because we would keep an invalid block ID if one day this
translated block is removed independently. Also, we'll have a consistent
way to convert the block accross all posts.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6565

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,8 @@
   },
   "require": {
     "roave/security-advisories": "dev-master",
-    "jakeasmith/http_build_url": "^1.0"
+    "jakeasmith/http_build_url": "^1.0",
+    "tightenco/collect": "5.3.20"
   },
   "require-dev": {
     "phpunit/phpunit": "~5.7",
--- a/src/class-integration-composite.php
+++ b/src/class-integration-composite.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace WPML\PB\Gutenberg;
+
+class Integration_Composite implements Integration {
+
+	/**
+	 * @var Integration[] $integrations
+	 */
+	private $integrations;
+
+	public function add( Integration $integration ) {
+		$this->integrations[] = $integration;
+	}
+	public function add_hooks() {
+		foreach ( $this->integrations as $integration ) {
+			$integration->add_hooks();
+		}
+	}
+
+}
--- a/src/reusable-blocks/class-reusable-blocks-integration.php
+++ b/src/reusable-blocks/class-reusable-blocks-integration.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace WPML\PB\Gutenberg;
+
+class Reusable_Blocks_Integration implements Integration{
+
+	/** @var Reusable_Blocks_Translation $reusable_blocks_translation */
+	private $reusable_blocks_translation;
+
+	public function __construct(
+		Reusable_Blocks_Translation $reusable_blocks_translation
+	) {
+		$this->reusable_blocks_translation = $reusable_blocks_translation;
+	}
+
+	public function add_hooks() {
+		add_filter( 'render_block_data', [ $this, 'convert_reusable_block' ] );
+	}
+
+	/**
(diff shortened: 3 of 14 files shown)


src/reusable-blocks/class-reusable-blocks.php
```diff
@@ -15,11 +15,7 @@ public function get_ids_from_post( $post_id ) {
 		if ( $post ) {
 			$blocks = \collect( \WPML_Gutenberg_Integration::parse_blocks( $post->post_content ) );
 
-			return $blocks->filter( function( $block ) {
-				return 'core/block' === $block['blockName']
-				       && isset( $block['attrs']['ref'] )
-				       && is_numeric( $block['attrs']['ref'] );
-			})->map( function( $block ) {
+			return $blocks->filter( [ __CLASS__, 'is_reusable' ] )->map( function( $block ) {
```
brucepearson: @strategio You may be able to write it like this. I'm not sure if you'll need the `flattern`
```
return $blocks->filter( [ __CLASS__, 'is_reusable' ] )
                       ->pluck( 'attrs' )
                       ->flattern()
                       ->pluck( 'ref' )
                       ->toArray();

```

brucepearson (approved):
@strategio Looks good. +1

andreasciamanna (approved):

strategio:
@brucepearson, @andreasciamanna, as we changed the workflow to handle reusable block translation, I closed https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/pull/32 and https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/pull/33. So I had to change the target branch and refactor the commit to include shared code from the previous ones. Please give it a look again.

composer.json
```diff
@@ -16,7 +16,8 @@
   },
   "require": {
     "roave/security-advisories": "dev-master",
-    "jakeasmith/http_build_url": "^1.0"
+    "jakeasmith/http_build_url": "^1.0",
+    "tightenco/collect": "5.3.20"
```
strategio, line 20: I will use this in the next task, so I kept this.

src/class-integration-composite.php
```diff
@@ -0,0 +1,30 @@
...
+	 * @param Integration[] $integrations
+	 */
+	public function __construct( array $integrations ) {
+		$this->integrations = $integrations;
+	}
+
+	public function add_hooks() {
+		foreach ( $this->integrations as $integration ) {
+
+			if ( ! $integration instanceof Integration ) {
```
brucepearson: @strategio I know this is strictly correct but throwing exceptions for this kind of things is redundant to me.

If you really want to make sure the class is used the correct way then I think you should not allow the class to except an array of unknown types in the first place but make the caller add the correct items.

eg.
```
class Integration_Composite implements Integration {
	/**
	 * @var Integration[] $integrations
	 */
	private $integrations = [];

	public function add( Integration $integration ) {
		$this->integrations[] = $integrations;
	}
 
...
}
```

This way the IDE will tell you're if you doing it wrong.

src/reusable-blocks/class-reusable-blocks-integration.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace WPML\PB\Gutenberg;
+
+class Reusable_Blocks_Integration implements Integration{
+
+	const POST_TYPE = 'wp_block';
```
brucepearson: @strategio This const is not used in this class so it probably belongs somewhere else. I would move to the `Reusable_Blocks_Translation` where it's used for now.

brucepearson (changes requested):

src/class-integration-composite.php
strategio: done @brucepearson, I opted for the `add` method.

brucepearson (approved):
@strategio Looks good. +1

PR #35, opened 2019-05-07
Automatically add jobs for reusable blocks

strategio:
We will scan the translation batch and maybe add the required reusable
elements.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6580


src/reusable-blocks/class-reusable-blocks-batch-handler.php
```diff
@@ -0,0 +1,114 @@
...
+	 * @return array
+	 */
+	public function select_target_langs( array $block ) {
+		foreach ( array_keys( $block['target_langs'] ) as $target_lang ) {
+			if ( ! $this->needs_job( $block['block_id'], $target_lang ) ) {
+				unset( $block['target_langs'][ $target_lang ] );
+			}
+		}
+
+		return $block;
```
brucepearson: @strategio I would make this more declarative:

```
$block['target_langs] = collect( $block['target_langs'] )
   ->filter( function ( $item, $target_lang )  use ( $block ) {
        return $this->requires_translation( $block['block_id'], $target_lang );
   } )
  ->toArray();
```

src/reusable-blocks/class-reusable-blocks-batch-handler.php
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	/**
+	 * @param int    $block_id
+	 * @param string $target_lang
+	 *
+	 * @return bool
+	 */
+	private function needs_job( $block_id, $target_lang ) {
```
brucepearson: @strategio I think a better name is `requires_translation`

src/reusable-blocks/class-reusable-blocks-translation.php
```diff
@@ -24,14 +24,20 @@ public function __construct( \SitePress $sitepress ) {
...
 	}
+
+	/**
+	 * @param int         $block_id
+	 * @param bool        $original_if_missing
+	 * @param string|null $lang
+	 *
+	 * @return
+	 */
+	public function convert_block_id( $block_id, $original_if_missing, $lang = null ) {
```
brucepearson: @strategio I would remove `$original_if_missing` because I think you always want to call it with true.

brucepearson (approved):
@strategio I made a few suggestions but it looks good. +1

src/reusable-blocks/class-reusable-blocks-batch-handler.php
strategio: @brucepearson, that's part of things I am not totally convinced when we do it in the functional way. If I want to use `requires_translation` inside `filter`, I have 2 options:

1. Pass `$this` to the closure via `use` (that's missing in your suggestion). But then I am not sure this is really correct to expose the whole object in the closure.

2. Previously map the target languages into a new entity which would contain all the information needed in the callback. That way (with only one argument), I can define the filter as a public method of the class: `->filter( [ $this, 'filter_target_langs'] )`. But then later, I need to re-map the target langs and remove the extra info.

I used the method (2) in some places when I thought it would ease the readability. But still, I think I need more practice to find the best approach.

I will try your suggestion though.

src/reusable-blocks/class-reusable-blocks-translation.php
strategio: @brucepearson, the method `convert_block` is calling it with `true`. That's for the frontend, so we display the original block if it's not translated.

src/reusable-blocks/class-reusable-blocks-batch-handler.php
strategio: I applied your suggestion in https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/pull/35/commits/177c348aebcdfa462644446d4a502216ab0d3e81. Not sure this is cleaner than the previous version, but it passes the tests. Let me know what you think?

strategio:
@OnTheGoSystems/wpml

src/reusable-blocks/class-reusable-blocks-batch-handler.php
brucepearson: @strategio $this should be available inside a closure since php 5.4. https://php.net/manual/en/functions.anonymous.php#example-169

src/reusable-blocks/class-reusable-blocks-translation.php
brucepearson: @strategio I'm suggesting to remove it as a parameter and change the following line to:

```
return $this->sitepress->get_object_id( $block_id, self::POST_TYPE, true, $lang );
```

src/reusable-blocks/class-reusable-blocks-batch-handler.php
strategio: That's right @brucepearson! Corrected :)

src/reusable-blocks/class-reusable-blocks-batch-handler.php
```diff
@@ -0,0 +1,114 @@
...
+		 *      'block_id'     => 1,
+		 *      'target_langs' => ['fr' => 1, 'de' => 1],
+		 *      'source_lang'  => 'en',
+		 *  ],
+		 *  [
+		 *      'block_id'     => 2,
+		 *      'target_langs' => ['de' => 1],
+		 *      'source_lang'  => 'en',
+		 *  ],
+		 * ]
```
andreasciamanna: @strategio does it have to be an array of associative arrays?

Can't we make it an array of a specific [plain old object](https://en.wikipedia.org/wiki/Plain_old_Java_object)?

Please don't bring the "getters/setters" issues: the point of this kind of objects is that they are as simple as possible, but with all the advantages of dealing with a typed object (versus an undetermined associative array).

Below, you have this:

```php
$new_element = new \WPML_TM_Translation_Batch_Element(
	$block['block_id'],
	'post',
	$block['source_lang'],
	$block['target_langs']
);
```

That's a classic case of code prone to bugs caused by typos or not properly planned change in arrays keys (not that uncommon, looking at the recent issues we released to clients).

This would be less prone to bugs because the code editor (a properly configured one, that is) would show if we are using a property of the object which doesn't exist or that returns the wrong data:

```php
$new_element = new \WPML_TM_Translation_Batch_Element(
	$block->block_id,
	'post',
	$block->source_lang,
	$block->target_langs
);
```

andreasciamanna (changes requested):
The PR looks fine.

However, I don't find the use of associative arrays to define a block the best option.

src/reusable-blocks/class-reusable-blocks-batch-handler.php
strategio: I will do this in the next PR https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/pull/37 if you don't have objection @andreasciamanna (it will be easier since I refactored a good part of the code).

andreasciamanna: 👍

andreasciamanna (approved):

src/reusable-blocks/class-reusable-blocks-batch-handler.php
strategio: Done here https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/pull/37/commits/941bfdcbb3b19acd042ff264e43fa44d5c40a68d.

andreasciamanna: 🎉

PR #37, opened 2019-05-09
Automatically add reusable blocks in the translation basket

strategio:
As the user has the ability to manually remove items in the translation
basket, we will not automatically add the blocks in the batch when the
basket is submitted.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6590


src/reusable-blocks/class-reusable-blocks-basket-element.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace WPML\PB\Gutenberg;
+
+class Reusable_Blocks_Basket_Element {
```
strategio: We don't have a specific object to hold the basket items. I created this class which is quite similar to `\WPML_TM_Translation_Batch_Element` and allows to share a good part of the code.

src/reusable-blocks/class-reusable-blocks-basket-handler.php
```diff
@@ -0,0 +1,82 @@
...
+			->toArray();
+
+		if ( $blocks ) {
+			$basket_portion = [
+				'post'              => [],
+				'source_language'	=> $data['translate_from'],
+				'target_languages'	=> array_keys( $data['tr_action'] ),
+			];
+
+			foreach ( $blocks as $block ) {
```
strategio: I started writing this piece of code in the functional way, but there's a bug with `mapWithKeys` in version 5.3 (https://github.com/laravel/framework/issues/15409), it won't work with numeric keys.

strategio:
@OnTheGoSystems/wpml please review after https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/pull/35

src/reusable-blocks/class-reusable-blocks-basket-handler.php
```diff
@@ -28,7 +28,7 @@ public function add_blocks( array $data ) {
 		$blocks        = $this->get_blocks_from_post_elements( $post_elements );
 
 		$blocks = $blocks->map( [ $this, 'select_target_langs' ] )
-			->filter( function( $block ) { return ! empty( $block['target_langs'] ); } )
+			->filter( function( $block ) { return ! empty( $block->target_langs ); } )
```
brucepearson: @strategio You can use reject instead of the not empty. It reads nicer to me.
```php
->reject( function( $block ) { return empty( $block->target_langs ); } )
```

brucepearson (approved):
@strategio Looks good. +1

src/reusable-blocks/class-reusable-blocks-basket-handler.php
strategio: Good idea @brucepearson! New habits are coming :)

src/class-wpml-gutenberg-integration-factory.php
```diff
@@ -41,8 +42,13 @@ public function create() {
...
 				new WPML\PB\Gutenberg\Reusable_Blocks_Admin_Integration(
 					new WPML\PB\Gutenberg\Reusable_Blocks_Batch_Handler(
-						new WPML\PB\Gutenberg\Reusable_Blocks(),
+						$reusable_blocks,
 						$reusable_blocks_translation
+					),
+					new WPML\PB\Gutenberg\Reusable_Blocks_Basket_Handler(
+						$reusable_blocks,
+						$reusable_blocks_translation,
+						new \WPML_Translation_Basket( $wpdb )
```
andreasciamanna: @strategio We could use the DI (see what I did here? ;) ) helper for creating the instance of `\WPML_Translation_Basket`.

That would help to test this code as well.

src/reusable-blocks/class-reusable-blocks-basket-element.php
andreasciamanna: It might be a good idea to extract the constructor and the getters into a parent (abstract, if needed) class so that both `\WPML_TM_Translation_Batch_Element` and ´\WPML\PB\Gutenberg\Reusable_Blocks_Basket_Element´ would share the code instead of duplicating it.

src/reusable-blocks/class-reusable-blocks-basket-handler.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+namespace WPML\PB\Gutenberg;
+
+class Reusable_Blocks_Basket_Handler extends Reusable_Blocks_Handler {
```
andreasciamanna: I don't like the use of the word "Handler".

Sometimes we can't avoid that, but whenever I can, I tend to follow [these advises](https://www.yegor256.com/2015/03/09/objects-end-with-er.html).

In this case, can't we call this class just `Reusable_Blocks_Basket`?

Also, didn't we agree to use cameCase for class and method names when they are under a namespace?

In fact, I think we should have a `\WPML\PB\Gutenberg\ReusableBlocks` namespace under which we have this class.

That would make easy to name this class something like `ManageBasket`, for instance. Or `HandleBasker`.  
Or even `BasketHandler` is we can't find a better name.

andreasciamanna (changes requested):
@strategio a summary of my inline comments:

- DI helpers (see what I did here? ;) ) instead of hard-dependencies
- We agreed on using camelCase convention for classes and methods which are under a namespace and for namespaces as well
- If possible, "Don't Create Objects That End With -ER"

src/reusable-blocks/class-reusable-blocks-basket-element.php
strategio: @andreasciamanna, at the beginning I extended `Reusable_Blocks_Basket_Element` from `\WPML_TM_Translation_Batch_Element` but this was a bit hacky. Then later I add troubles with tests because `\WPML_TM_Translation_Batch_Element` is in another project.

If I start sharing code for this simple class, I will still have to create a stub in this project for testing. That's why I opted for this solution.

src/reusable-blocks/class-reusable-blocks-basket-handler.php
andreasciamanna: Ok.

We can address that in the future then.

src/reusable-blocks/class-reusable-blocks-basket-element.php
andreasciamanna: Ok.

We can address that in the future then.

src/reusable-blocks/class-reusable-blocks-basket-handler.php
strategio: At the time I started this project, we didn't agreed with the new format yet. I will add a commit to refactor all classes related to reusable blocks.

I renamed the classes and methods + extended the namespace => https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/pull/37/commits/39698cb5ac7e2e040004e3b84b46d321a4ca7c19

src/class-wpml-gutenberg-integration-factory.php
strategio: Done @andreasciamanna https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/pull/37/commits/92d5d7397ad5a713d218e9e6f2e880a6671698f6

andreasciamanna (approved):

src/class-wpml-gutenberg-integration-factory.php
```diff
@@ -31,18 +31,24 @@ public function create() {
...
-						new WPML\PB\Gutenberg\Reusable_Blocks(),
+				new WPML\PB\Gutenberg\ReusableBlocks\AdminIntegration(
+					new WPML\PB\Gutenberg\ReusableBlocks\ManageBatch(
+						$reusable_blocks,
 						$reusable_blocks_translation
+					),
+					new WPML\PB\Gutenberg\ReusableBlocks\ManageBasket(
+						$reusable_blocks,
+						$reusable_blocks_translation,
+						\WPML\Container\make( '\WPML_Translation_Basket' )
```
brucepearson, line 51: @strategio Since you are using DIC here, I think we should use it correctly.

This whole code can be replaced with:
```
$integrations->add( \WPML\Container\make( 'WPML\PB\Gutenberg\ReusableBlocks\Integration' );
if ( is_admin() ) {
   $integrations->add( \WPML\Container\make( 'WPML\PB\Gutenberg\ReusableBlocks\AdminIntegration' );
}
```
BTW, What happens if TM is not active and `WPML_Translation_Basket` doesn't exist? ( Maybe this code only runs when TM is active, I don't remember )

strategio: @brucepearson, I implemented your suggestion in https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/pull/38.

PR #38, opened 2019-05-13
Do not load reusable block integration when it's not translatable

strategio:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6601

The second commit is to address a suggestion made by @brucepearson in https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/pull/37#discussion_r283064217.

### DIFF
--- a/src/class-wpml-gutenberg-integration-factory.php
+++ b/src/class-wpml-gutenberg-integration-factory.php
@@ -31,27 +31,33 @@ class WPML_Gutenberg_Integration_Factory {
 			)
 		);
 
-		$reusable_blocks             = new WPML\PB\Gutenberg\ReusableBlocks\Blocks();
-		$reusable_blocks_translation = new WPML\PB\Gutenberg\ReusableBlocks\Translation( $sitepress );
-
-		$integrations->add(
-			new WPML\PB\Gutenberg\ReusableBlocks\Integration( $reusable_blocks_translation )
+		$is_reusable_block_translatable = $sitepress->is_translated_post_type(
+			WPML\PB\Gutenberg\ReusableBlocks\Translation::POST_TYPE
 		);
 
-		if ( is_admin() ) {
+		if ( $is_reusable_block_translatable ) {
+			$reusable_blocks             = new WPML\PB\Gutenberg\ReusableBlocks\Blocks();
+			$reusable_blocks_translation = new WPML\PB\Gutenberg\ReusableBlocks\Translation( $sitepress );
+
 			$integrations->add(
-				new WPML\PB\Gutenberg\ReusableBlocks\AdminIntegration(
-					new WPML\PB\Gutenberg\ReusableBlocks\ManageBatch(
-						$reusable_blocks,
-						$reusable_blocks_translation
-					),
-					new WPML\PB\Gutenberg\ReusableBlocks\ManageBasket(
-						$reusable_blocks,
-						$reusable_blocks_translation,
-						\WPML\Container\make( '\WPML_Translation_Basket' )
-					)
-				)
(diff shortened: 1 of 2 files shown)


strategio:

src/class-wpml-gutenberg-integration-factory.php
```diff
@@ -31,27 +31,20 @@ public function create() {
...
-
-		$integrations->add(
-			new WPML\PB\Gutenberg\ReusableBlocks\Integration( $reusable_blocks_translation )
+		$is_reusable_block_translatable = $sitepress->is_translated_post_type(
+			WPML\PB\Gutenberg\ReusableBlocks\Translation::POST_TYPE
 		);
 
-		if ( is_admin() ) {
+		if ( $is_reusable_block_translatable ) {
 			$integrations->add(
```
brucepearson, line 35: @strategio Do you need to share `WPML\PB\Gutenberg\ReusableBlocks\Blocks` and `WPML\PB\Gutenberg\ReusableBlocks\Translation` like you do in the original code?

eg.
```php
WPML\Container\share( [
    'WPML\PB\Gutenberg\ReusableBlocks\Blocks',
   'WPML\PB\Gutenberg\ReusableBlocks\Translation',
]);
```

src/class-wpml-gutenberg-integration-factory.php
```diff
@@ -31,27 +31,20 @@ public function create() {
...
-		$reusable_blocks_translation = new WPML\PB\Gutenberg\ReusableBlocks\Translation( $sitepress );
-
-		$integrations->add(
-			new WPML\PB\Gutenberg\ReusableBlocks\Integration( $reusable_blocks_translation )
+		$is_reusable_block_translatable = $sitepress->is_translated_post_type(
+			WPML\PB\Gutenberg\ReusableBlocks\Translation::POST_TYPE
 		);
 
-		if ( is_admin() ) {
+		if ( $is_reusable_block_translatable ) {
```
brucepearson: @strategio I think it would be better:

`if( $this->should_translate_reusable_blocks() ) {`

brucepearson (approved):
Some minor comments. +1

src/class-wpml-gutenberg-integration-factory.php
strategio: @brucepearson, sharing is not really needed because the objects don't have any specific state. But we can save some memory (and I keep forgetting to use this `share` function). That's fixed!

src/class-wpml-gutenberg-integration-factory.php
strategio: Good idea @brucepearson, done!

strategio:
@OnTheGoSystems/wpml, please one more review!

src/class-wpml-gutenberg-integration-factory.php
brucepearson: @strategio I only mentioned it so you are aware of it. I only recommend sharing when it's required and not just to save a few bytes. Now it could be misleading because someone reading the code might assume that it's a requirement that there's only ever one instance of these classes.

strategio: @brucepearson, in some DIC the default is that instances are shared.

Do you think that we should **not share** when it's not explicitly required? I can remove it if you prefer.

strategio:
@OnTheGoSystems/wpml please one more review!

src/class-wpml-gutenberg-integration-factory.php
strategio: Removed the class instance sharing!

andreasciamanna (approved):

PR #40, opened 2019-05-16
Register/update attribute strings according to the config

strategio:
I reorganized the code to separate the parsing of HTML (the existing code) and attributes (the new code).
This is now grouped into a composite class.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6607


src/strings-in-block/class-html.php
```diff
@@ -0,0 +1,190 @@
+<?php
+
+namespace WPML\PB\Gutenberg\StringsInBlock;
+
+class HTML extends Base {
```
strategio, line 5: This class is the transposition of the old class `WPML_Gutenberg_Strings_In_Block`. And I moved some shared code inside the `Base` class.

strategio:

src/strings-in-block/class-attributes.php
```diff
@@ -0,0 +1,159 @@
+<?php
+
+namespace WPML\PB\Gutenberg\StringsInBlock;
+
+class Attributes extends Base {
```
andreasciamanna, line 5: @strategio (not blocking) I suggest adjusting the name and location of the new files:

- the path of the file must meet the namespace. In this case, it should be `src/WPML/PB/Gutenberg/StringsInBlock` (*)
- the filename must be the same as the class. In this case, `Attributes .php`
- In other words, the full path would be `src/WPML/PB/Gutenberg/StringsInBlock/Attributes .php` (*)

We can even decide that `src` represent our root namespace (`WPML`): as far as I've seen, it's a quite common approach.

You can even tell PHPStorm that the root namespace is under `src/WPML` (or `src`) so that new classes are created accordingly.
For instance, you can create a class `Foo` under the `WPML/Bar` namespace and PHPStorm will take care of creating it in the right place and with the right name.

src/strings-in-block/class-composite.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+namespace WPML\PB\Gutenberg\StringsInBlock;
+
+class Composite implements StringsInBlock {
```
andreasciamanna: @strategio I don't get the meaning of `Composite`

andreasciamanna (changes requested):

src/strings-in-block/class-composite.php
strategio: It's the [composite pattern](https://sourcemaking.com/design_patterns/composite) @andreasciamanna. I don't have a better idea to name this...

src/strings-in-block/class-attributes.php
strategio: Honnestly, I am not convinced in this context @andreasciamanna. Maybe we should have used a namespace dedicated to this project making abstration of the parent project (i.e. `WPMLGutenberg`). But now I have another feature branch (reusable blocks) almost ready to be merged and it will use the long namespace with the current file structure.

src/strings-in-block/class-composite.php
andreasciamanna: Unless it represents an entity, I usually prefer using verbs or adjectives.

What about "Compose"?

That's the purpose of the composite pattern after all: to **compose** different entities into something which can be handled as a whole.

src/strings-in-block/class-attributes.php
andreasciamanna: @kkarpieszuk I don't know which part of what I wrote does not convince you.

Also, I got lost after "abstraction"... we can discuss this by voice next week.

For now, I think we can survive with what we have, but we must better define (with the rest of the team and in written form) a convention for namespaces, naming and location of the files.

Or it will be soon become a nightmare.

src/strings-in-block/class-composite.php
strategio: Done!

src/strings-in-block/class-attributes.php
strategio: I agree @andreasciamanna. I can adjust the namespaces and file strucutre later when we decided.

strategio:
@andreasciamanna, I added 2 commits to improve the feature:
1. f292671 FIXUP Add support for regex keys: We automatically detect regex keys and use it.
2. 6b30861 FIXUP Do not register numbers.

tests/phpunit/tests/strings-in-block/test-attributes.php
```diff
@@ -0,0 +1,285 @@
...
+ * @group strings-in-block
+ */
+class TestAttributes extends \OTGS_TestCase {
+
+	const BLOCK_NAME = 'block/type';
+
+	/**
+	 * @test
+	 */
+	public function it_should_find() {
```
andreasciamanna: @strategio It should find what? Please be specific when naming tests.

andreasciamanna (changes requested):
@strategio You already know I don't like this approach because we discussed that.

I believe that making our code guess if the key is using wildcards or a regular expression is an excellent exercise but not something we should have in production.

I still believe what I suggested during our chat is a better approach: define an optional ”search-method” attribute in the XSD for the ”key” element (*).
It defaults to ”wildcards” if omitted. If provided, it will accept either ”wildcards” or ”regex”.

If parsIng the key elements with the new attribute is too complicated, what about adding the new attribute to the parent element (gutenberg-block)?
Would that make things easier? 
In this case, the search method would be global for the current block, it can be still defined as I shown to you via Mattermost, and we can reuse it for other element types in the future, keeping it both backwards compatible and consistent across all the configuration elements.

If a new solution is preferable, rewriting the code is acceptable and recommended.

Please remember that the regular expression is feature which wasn't even part of the specs. If we add it, I want it to be done right from the start.

tests/phpunit/tests/strings-in-block/test-attributes.php
strategio: @andreasciamanna, I created another PR https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/pull/41 to introduce the parameter `search-method`.

strategio: Also fixed the test method names.

andreasciamanna: @strategio it's still called `it_should_find`, isn't it?

andreasciamanna (approved):
@strategio the ambiguous `it_should_find` is still there.

However, I was more focused on the way you decide if we should use wildcards or regular expressions. You handled it in the other PR, which looks good to me.

tests/phpunit/tests/strings-in-block/test-attributes.php
strategio: It's fixed in the other PR Andrea.

strategio:
@OnTheGoSystems/wpml one more review please.

src/strings-in-block/class-compose.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+namespace WPML\PB\Gutenberg\StringsInBlock;
+
+class Compose implements StringsInBlock {
```
brucepearson: @strategio `Compose` doesn't seem to be a meaningful name to me. I assume this is the `Composite` pattern so I would use `Composite` for this.

src/strings-in-block/class-compose.php
```diff
@@ -0,0 +1,64 @@
...
+	 * @param \WP_Block_Parser_Block $block
+	 *
+	 * @return array
+	 */
+	public function find( \WP_Block_Parser_Block $block ) {
+		$strings = [];
+
+		foreach ( $this->parsers as $parser ) {
+			$strings = array_merge( $strings, $parser->find( $block ) );
+		}
```
brucepearson: @strategio If $this->parsers is made as a collection in the constructor then you can do the following. But I'm not sure if it's better or not.

```php
$strings = $this->parsers->reduce( 
   function( $strings, $parser ) use ( $block ) {
      return $strings->merge( $parser->find( $block ) ); 
  },
  collect( [] )
)->toArray();
```

src/strings-in-block/class-compose.php
```diff
@@ -0,0 +1,64 @@
...
+		 *
+		 * When post with Gutenberg blocks is being send for translation
+		 * WPML parses blocks to find translatable strings. With this filter
+		 * you can register additional strings to be added to Translation
+		 * Package and then translated
+		 *
+		 * @param array                  $strings already found strings.
+		 * @param \WP_Block_Parser_Block $block block being parsed.
+		 */
+		return apply_filters( 'wpml_found_strings_in_block', $strings, $block );
```
brucepearson: @strategio I'm not sure if this should be here. To me it seems like another thing that could implement `StringsInBlock`. 
Then the factory would change to:
```php
$string_parsers = [
			new WPML\PB\Gutenberg\StringsInBlock\HTML( $config_option ),
			new WPML\PB\Gutenberg\StringsInBlock\Attributes( $config_option ),
			new WPML\PB\Gutenberg\StringsInBlock\APIFilters(),
		];
```

You'll need to change the signature of the `find` function to also receive the strings.

This will then keep everything separate.

src/strings-in-block/class-compose.php
andreasciamanna: I suggested using "Compose" because:

> That's the purpose of the composite pattern after all: to compose different entities into something which can be handled as a whole.

It still does not convince me, as I don't think the method should tell the reader that "I'm implementing the Composite pattern".

In general, I'm not a big fan of class names telling you I'm implementing a specific pattern: the class should represent an entity or a functionality: if we replace the pattern we use, we will need to rename all these classes.

On the other hand, I realise now that all factory classes contain the word "Factory", which would go in favour of using the word "Composite" here. At least for consistency.

So I'm fine to rollback to "Composite" only for that reason.

brucepearson: Maybe we should name if for what it is. `Parsers` or `ParserCollection` or just `Collection`

src/strings-in-block/class-compose.php
strategio: @brucepearson, I understand your suggestion but it does not guarantee that the `APIFilters` parser will be at the end (and the order is not really testable in the factory).

The order is important because I have one case where we need to remove a string from the found collection (https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6611).

src/strings-in-block/class-compose.php
strategio: The working code is:

```php
// With $parsers collected in the constructor.
$strings = $this->parsers->reduce(
	function( $strings, $parser ) use ( $block ) {
		return collect( $strings )->merge( collect( $parser->find( $block ) ) );
	},
	collect( [] )
)->toArray();
```

To me it seems harder to read than:

```php
foreach ( $this->parsers as $parser ) {
	$strings = array_merge( $strings, $parser->find( $block ) );
}
```

Maybe because I am not so used to this paradigm.

src/strings-in-block/class-compose.php
strategio: I will go for `Collection` then :)

src/strings-in-block/class-compose.php
brucepearson: Yes, keep the simpler version. But I'm surprised you needed those extra collect because `$strings` in the function should have already been a collection.

brucepearson (approved):
@strategio Just change the class name from `Compose` to `Collection` and it looks good

src/strings-in-block/class-compose.php
strategio: Well, if we really implement the composite pattern, the `HTML` and `Attributes` classes should have the same method signatures as `Collection`. For `find` we are supposed to return an `array`.

strategio:
@brucepearson, could you also have a look at https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/pull/41.

I will also replace the attribute `key` with `attr` in https://git.onthegosystems.com/wpml/sitepress-multilingual-cms/merge_requests/2384#note_169998 (couldn't do this yesterday).

PR #42, opened 2019-05-20
Add support for nested reusable blocks translation

strategio:
`Blocks::getIdsFromPost` will lookup recursively in blocks.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6598

### DIFF
--- a/src/reusable-blocks/class-blocks.php
+++ b/src/reusable-blocks/class-blocks.php
@@ -16,11 +16,14 @@ class Blocks {
 	}
 
 	/**
+	 * We get block IDs recursively to find possible
+	 * nested reusable blocks.
+	 * 
 	 * @param int $post_id
 	 *
 	 * @return array
 	 */
-	public function getIdsFromPost( $post_id ) {
+	public function getChildrenIdsFromPost( $post_id ) {
 		$post = get_post( $post_id );
 
 		if ( $post ) {
@@ -30,8 +33,9 @@ class Blocks {
 				       && isset( $block['attrs']['ref'] )
 				       && is_numeric( $block['attrs']['ref'] );
 			})->map( function( $block ) {
-				return (int) $block['attrs']['ref'];
-			})->toArray();
+				$block_id = (int) $block['attrs']['ref'];
+				return array_merge( [ $block_id ], $this->getChildrenIdsFromPost( $block_id ) );
+			})->flatten()->toArray();
 		}
 
 		return [];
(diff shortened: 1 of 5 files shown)


strategio:
@OnTheGoSystems/wpml please review

src/reusable-blocks/class-blocks.php
```diff
@@ -16,6 +16,9 @@ public static function isReusable( array $block ) {
 	}
 
 	/**
+	 * We get block IDs recursively to find possible
+	 * nested reusable blocks.
+	 *
```
andreasciamanna, line 21: @strategio I think the name of the method is a bit unclear. What about `getChildrenIdsFromPost` or something like that?

src/reusable-blocks/class-blocks.php
```diff
@@ -30,8 +33,9 @@ public function getIdsFromPost( $post_id ) {
 				       && isset( $block['attrs']['ref'] )
 				       && is_numeric( $block['attrs']['ref'] );
 			})->map( function( $block ) {
-				return (int) $block['attrs']['ref'];
-			})->toArray();
+				$block_id = (int) $block['attrs']['ref'];
+				return array_merge( [ $block_id ], $this->getIdsFromPost( $block_id ) );
```
andreasciamanna: @strategio I think the name of the method is a bit unclear. 

What about `getChildrenIdsFromPost` or something like that?

It would make clear to the reader that you are adding the IDs of the nested blocks.

andreasciamanna (changes requested):

src/reusable-blocks/class-blocks.php
strategio: Done @andreasciamanna

src/reusable-blocks/class-blocks.php
strategio: Done!

andreasciamanna (approved):

tests/phpunit/tests/reusable-blocks/test-blocks.php
```diff
@@ -105,10 +105,96 @@ public function it_should_get_ids_from_post() {
...
+	 * @group wpmlcore-6598
+	 */
+	public function it_should_get_nested_block_ids_from_post() {
+		$GLOBALS['wp_version'] = '5.1.0';
+
+		$post_id        = 123;
+		$block_id       = 456;
+		$child_block_id = 789;
+
+		// First stack
```
brucepearson: @strategio I have no idea what `First stack`, `Second stack` and `Last stack` mean. Maybe you need to move the code for each part into a member function with a more meaningful name

strategio: Done @brucepearson, I refactored the expectations into `set_block_in_post`, easier to read now.

brucepearson (approved):
Thanks @strategio +1

PR #43, opened 2019-05-23
Allow to define a global configuration for a block namespace

strategio:
If no config is found for the specific block, we'll try to find a
configuration for the block namespace.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6635

I also added a second commit to make sure `blockName` attribute is set. I never got this situation in my tests, but I saw that we have this kind of check for the HMTL strings, so let's have the same verification for Attributes.

### DIFF
--- a/src/strings-in-block/class-attributes.php
+++ b/src/strings-in-block/class-attributes.php
@@ -178,7 +178,7 @@ class Attributes extends Base {
 	 * @return array
 	 */
 	private function getAttributes( \WP_Block_Parser_Block $block ) {
-		return ! empty( $block->attrs ) && is_array( $block->attrs ) ? $block->attrs : [];
+		return is_array( $block->attrs ) && $block->blockName ? $block->attrs : [];
 	}
 
 	/**
--- a/src/strings-in-block/class-base.php
+++ b/src/strings-in-block/class-base.php
@@ -27,9 +27,18 @@ abstract class Base implements StringsInBlock {
 
 		if ( isset( $block->blockName, $this->block_types[ $block->blockName ][ $type ] ) ) {
 			return $this->block_types[ $block->blockName ][ $type ];
-		} else {
-			return null;
 		}
+
+		if ( isset( $block->blockName ) ) {
+			$block_name_arr  = explode( '/', $block->blockName );
+			$block_namespace = reset( $block_name_arr );
+
+			if ( isset( $this->block_types[ $block_namespace ][ $type ] ) ) {
+				return $this->block_types[ $block_namespace ][ $type ];
+			}
+		}
+
+		return null;
 	}
 
 	/**
--- a/tests/phpunit/tests/strings-in-block/test-attributes.php
+++ b/tests/phpunit/tests/strings-in-block/test-attributes.php
@@ -9,7 +9,29 @@ namespace WPML\PB\Gutenberg\StringsInBlock;
  */
 class TestAttributes extends \OTGS_TestCase {
 
-	const BLOCK_NAME = 'block/type';
+	const BLOCK_NAMESPACE = 'block';
+	const BLOCK_NAME      = 'block/type';
+
+	/**
+	 * @test
+	 */
+	public function it_should_not_find_attribute_strings_if_blockname_is_not_set() {
+		$config_array = [
+			self::BLOCK_NAME => [
+				'key' => [],
+			],
+		];
+
+		$block = $this->getBlock();
+		$block->attrs = [ 'foo' => 'bar' ];
+
+		$config  = $this->getConfig( $config_array );
+		$subject = $this->getSubject( $config );
+
+		$strings = $subject->find( $block );
+
+		$this->assertEmpty( $strings );
+	}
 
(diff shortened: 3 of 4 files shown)


strategio:

kagg-design:
+1 @strategio

strategio:
@kagg-design, thumb up?

kagg-design (approved):
+1 @strategio

kagg-design:
> @kagg-design, thumb up?

Ooops )

strategio:
One more review please @OnTheGoSystems/wpml

src/strings-in-block/class-attributes.php
```diff
@@ -13,7 +13,7 @@ public function find( \WP_Block_Parser_Block $block ) {
 		$strings = [];
 		$attrs   = $this->getAttributes( $block );
 
-		if ( $attrs ) {
+		if ( $attrs && isset( $block->blockName ) ) {
```
andreasciamanna: @strategio 

I have a couple of questions here.

1.

`$block` is an object of type `\WP_Block_Parser_Block` which has a field called `blockName`: how cannot be set? I find it quite impossible. Perhaps it can be an empty string?

2.

if you use `$attrs` only when `$block->blockName` is set (or probably only when is not falsy), perhaps you could move this check right into `$this->getAttributes( $block )` changing it from that:

```php
return ! empty( $block->attrs ) && is_array( $block->attrs ) ? $block->attrs : [];
```

To that:
```php
return ! empty( $block->attrs ) && is_array( $block->attrs ) && $block->blockName ? $block->attrs : [];
```

By the way, can you please avoid double negations like in `! empty( $block->attrs )`? You can just write `$block->attrs` to get the same result:

```php
return $block->attrs && is_array( $block->attrs ) && $block->blockName ? $block->attrs : [];
```

Moreover, `attrs` is also a field of `WP_Block_Parser_Block`. The expression can be further simplified as such:

```php
return is_array( $block->attrs ) && $block->blockName ? $block->attrs : [];
```

src/strings-in-block/class-base.php
```diff
@@ -27,9 +27,16 @@ protected function get_block_config( \WP_Block_Parser_Block $block, $type ) {
 
 		if ( isset( $block->blockName, $this->block_types[ $block->blockName ][ $type ] ) ) {
 			return $this->block_types[ $block->blockName ][ $type ];
-		} else {
-			return null;
+		} elseif ( isset( $block->blockName) ) {
```
andreasciamanna: @strategio you don't need the "else" here:

```php
protected function get_block_config( \WP_Block_Parser_Block $block, $type ) {
	if ( null === $this->block_types ) {
		$this->block_types = $this->config_option->get();
	}
	if ( isset( $block->blockName, $this->block_types[ $block->blockName ][ $type ] ) ) {
		return $this->block_types[ $block->blockName ][ $type ];
	} 
	
	if ( isset( $block->blockName) ) {
		$block_name_arr  = explode( '/', $block->blockName );
		$block_namespace = reset( $block_name_arr );
		if ( isset( $this->block_types[ $block_namespace ][ $type ] ) ) {
			return $this->block_types[ $block_namespace ][ $type ];
		}
	}
	return null;
}
```

src/strings-in-block/class-base.php
```diff
@@ -27,9 +27,16 @@ protected function get_block_config( \WP_Block_Parser_Block $block, $type ) {
...
+			$block_name_arr  = explode( '/', $block->blockName );
+			$block_namespace = reset( $block_name_arr );
+
+			if ( isset( $this->block_types[ $block_namespace ][ $type ] ) ) {
+				return $this->block_types[ $block_namespace ][ $type ];
+			}
 		}
+
+		return null;
```
andreasciamanna, line 41: @strategio can't you return an empty array so you always return one type of data?

andreasciamanna (changes requested):

src/strings-in-block/class-attributes.php
strategio: Right @andreasciamanna, I think I was induced into this with the existing code. Indeed `WP_Block_Parser_Block` has these properties defined.

src/strings-in-block/class-base.php
strategio: Yes, actually I was already returning an empty array on the new class `Attributes::getKeyConfig` but I didn't change it on the existing one `HTML`.

That's done now.

src/strings-in-block/class-base.php
strategio: Right, done!

src/strings-in-block/class-base.php
strategio: Eventually, I'll keep the existing code because it causes a test to fail on the `HTML` side (in `HTML::find` where we check `is_array`). I am not sure the test is correct but I prefer not to change it now.

strategio:
Suggestions applied @andreasciamanna

andreasciamanna (approved):

PR #46, opened 2019-05-28
Add a notice showing the reusable block jobs automatically created

strategio:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6648

### DIFF
--- a/src/class-wpml-gutenberg-integration-factory.php
+++ b/src/class-wpml-gutenberg-integration-factory.php
@@ -44,11 +44,6 @@ class WPML_Gutenberg_Integration_Factory {
 			);
 
 			if ( is_admin() ) {
-				/** @var \WPML_Translation_Job_Factory $wpml_translation_job_factory */
-				global $wpml_translation_job_factory;
-
-				WPML\Container\share( [ $wpml_translation_job_factory, wpml_get_admin_notices() ] );
-
 				$integrations->add(
 					WPML\Container\make( '\WPML\PB\Gutenberg\ReusableBlocks\AdminIntegration' )
 				);
--- a/src/class-wpml-gutenberg-integration-factory.php
+++ b/src/class-wpml-gutenberg-integration-factory.php
@@ -44,6 +44,11 @@ class WPML_Gutenberg_Integration_Factory {
 			);
 
 			if ( is_admin() ) {
+				/** @var \WPML_Translation_Job_Factory $wpml_translation_job_factory */
+				global $wpml_translation_job_factory;
+
+				WPML\Container\share( [ $wpml_translation_job_factory, wpml_get_admin_notices() ] );
+
 				$integrations->add(
 					WPML\Container\make( '\WPML\PB\Gutenberg\ReusableBlocks\AdminIntegration' )
 				);
--- a/src/reusable-blocks/class-manage-batch.php
+++ b/src/reusable-blocks/class-manage-batch.php
@@ -21,7 +21,7 @@ class ManageBatch extends Manage {
 
 		$blocks_to_add = $this->getBlockElementsToAdd( $blocks );
 
-		if ( $blocks_to_add->count() < 1 ) {
+		if ( $blocks_to_add->isEmpty() ) {
 			return $batch;
 		}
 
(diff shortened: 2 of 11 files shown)


strategio:
@OnTheGoSystems/wpml review please!

src/class-wpml-gutenberg-integration-factory.php
```diff
@@ -37,6 +37,11 @@ public function create() {
 			);
 
 			if ( is_admin() ) {
+				/** @var \WPML_Translation_Job_Factory $wpml_translation_job_factory */
+				global $wpml_translation_job_factory;
+
+				WPML\Container\share( [ $wpml_translation_job_factory, wpml_get_admin_notices() ] );
```
brucepearson: @strategio I think these should be shared at the point where they are created and not here.

src/reusable-blocks/class-manage-batch.php
```diff
@@ -4,10 +4,30 @@
...
+		$this->notice = $notice;
+	}
+
 	public function addBlocks( \WPML_TM_Translation_Batch $batch ) {
 		$blocks = $this->getBlocksFromPostElements( \collect( $batch->get_elements() ) );
 
-		$this->getBlockElementsToAdd( $blocks )->each(
+		$blocks_to_add = $this->getBlockElementsToAdd( $blocks );
+
+		if ( $blocks_to_add->count() < 1 ) {
```
brucepearson: @strategio Why `< 1`? Wouldn't `if ( $blocks_to_add->isEmpty() ) {` be better?

src/reusable-blocks/class-manage-batch.php
```diff
@@ -4,10 +4,30 @@
...
+			return $batch;
+		}
+
+		add_action( 'wpml_added_translation_jobs', [ $this, 'notifyExtraJobsToTranslator' ] );
```
brucepearson: @strategio Adding an action that then calls another function in this class is a type of spaghetti code. Don't we know already in this function?

Otherwise I think you move the hook to the `Notice` class and responds to the event without involving this class.

brucepearson (changes requested):

src/reusable-blocks/class-manage-batch.php
strategio: @brucepearson, we have 2 ways to add the blocks:
1. Automatically create jobs when we click on the "+" icon (we add the blocks to the batch).
2. Automatically add block elements to the basket (then the translation manager can decide to keep it or not).

We want this notice to show up only when we click on the "+" icon (1). That's why we have the method `notifyExtraJobsToTranslator` in this class.

I could add the hook in the `AdminIntegration` class and set a private property (i.e. `$has_block_jobs`) in `ManageBatch` to know if we need this notice or not.

In the end, the logic is almost the same: We have to hook to `wpml_added_translation_jobs` only when we have blocks added in this `ManageBatch::addBlocks`.

src/class-wpml-gutenberg-integration-factory.php
strategio: Done @brucepearson:
- https://git.onthegosystems.com/wpml/sitepress-multilingual-cms/merge_requests/2393
- https://git.onthegosystems.com/wpml/wpml-translation-management/merge_requests/1395

This also had the consequence to display the notice immediately in CTE.
After a discussion with Amir, we agreed to show it always in post edit
or post list to make it consistent with ATE. That's why I added the
`set_restrict_to_screen_ids`.

src/reusable-blocks/class-manage-batch.php
strategio: Nice one @brucepearson :)

src/reusable-blocks/class-manage-batch.php
strategio: Finally, I found a better way to do this @brucepearson. I think I wanted to overoptimize the code to add the hook only when we have blocks in a direct job. Now it's clearer at least.

brucepearson (approved):
@strategio It looks good. +1

strategio:
@OnTheGoSystems/wpml one more please :)

andreasciamanna (approved):

strategio:
@brucepearson, I extracted some code to `JobLinks` and reworked the tests :)

brucepearson (approved):
@strategio Geat! +1

PR #47, opened 2019-05-30
Translate inner content strings in a block with inner blocks

strategio:
Until now, we where handling block with inner blocks as pure layouts
(like columns for instance). The content around the inner blocks was not
supposed to be translated as it was only html tags for the layout as below:

```php
innerContent => [
    '<div class="column-one">',
    null,
    '</div><div class="column-two">',
    null,
    '</div>'
]
```

FTR, the `null` elements are replaced with the content of the inner blocks.

But now some blocks are also having translatable strings around the inner
blocks.

Here's an example of the `innerContent` that we have now:

```php
innerContent => [
    '<div><p>The title</p>',
    null,
    '\n\n',
    null,
    '</div>'
]
```

I could not use the `DOM` to parse the inner content because it's made of
invalid HTML extracts taken independently. Using the `DOM`
parser was fixing those extracts to a valid HTML which is not what we
want here.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6627

### DIFF
--- a/tests/phpunit/tests/test-wpml-gutenberg-integration.php
+++ b/tests/phpunit/tests/test-wpml-gutenberg-integration.php
@@ -538,6 +538,144 @@ class Test_WPML_Gutenberg_Integration extends OTGS_TestCase {
 
 	/**
 	 * @test
+	 * @group wpmlcore-6627
+	 */
+	public function it_updates_root_content_and_inner_blocks() {
+		$original_post               = \Mockery::mock( 'WP_Post' );
+		$original_post->post_content = 'Post content / Will be replaced with parsed blocks';
+
+		$translated_post_id = 123;
+		$target_lang        = 'de';
+
+		$parent_block_name      = 'some-block-name';
+		$core_parent_block_name = 'core/' . $parent_block_name; // Gutenberg prefixes with 'core/'
+		$p_block_name           = 'paragraph';
+		$core_p_block_name      = 'core/' . $p_block_name; // Gutenberg prefixes with 'core/'
+
+		$config_option = \Mockery::mock( 'WPML_Gutenberg_Config_Option' );
+		$config_option->shouldReceive( 'get' )
+		              ->andReturn(
+		              	[
+			                $core_parent_block_name => [
+			                	'xpath' => [
+			                	    '//summary',
+				                    '//details/@data-attr',
+			                    ]
+			                ],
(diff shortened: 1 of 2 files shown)


strategio:
@OnTheGoSystems/wpml review please :)

tests/phpunit/tests/test-wpml-gutenberg-integration.php
```diff
@@ -529,6 +529,144 @@ public function it_updates_inner_blocks_with_innerContent() {
 
 	}
 
+	/**
+	 * @test
+	 * @group wpmlcore-6627
+	 */
+	public function it_updates_root_content_and_inner_blocks() {
```
strategio, line 543: I wrote tests in this integration class otherwise we would miss the recursive part of the inner blocks replacement. I tried to detail test steps as much as I could, but we are testing many things at this point :)

andreasciamanna: Huge test: I'll have to trust you here :)

andreasciamanna (dismissed):

strategio:
@OnTheGoSystems/wpml please review :)

tests/phpunit/tests/test-wpml-gutenberg-integration.php
brucepearson: @strategio How about extracting `update_string_in_innerContent` into a separate class that you can inject into this class. Then you only need to test that it's called with the correct arguments. 
Then the function in the new class can be tested separately which should be much simpler.

brucepearson (changes requested):

tests/phpunit/tests/test-wpml-gutenberg-integration.php
strategio: @brucepearson, even if I extract `update_string_in_innerContent`, we would still have to test it from the integration class. The method I am testing here is `WPML_Gutenberg_Integration::string_translated` which is calling `WPML_Gutenberg_Integration::update_block_translations`, a recursive function.

The idea is not only to cover the code (which I could do by a test in `TestHTML`), but to test a real block with inner blocks. And that's why we need to test the recursion part here.

brucepearson: ok

brucepearson (dismissed):
@strategio +1

PR #48, opened 2019-05-31
Correctly set the restricted screen IDs for the reusable block jobs notice

strategio:
I wrongly thought that the screen ID was the same regardless the post
type. We will get the screen ID information from the `return_url`
parameter.

Also added a second commit for polishing (see screenshots below):
- Allow to hide the notice.
- Show the external link icon.

notice-in-editor

notice-in-list

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6652


strategio:
@OnTheGoSystems/wpml review please!

PR #50, opened 2019-05-31
Return an empty array when a block has an empty config

strategio:
It means that the block has nothing to translate. Instead, when we
return `null` for the HTML parsing, it means that the block has no
config and we will register the whole HTML of the block as a string.

We had a unit test on this and I mistakenly updated it while I refactored the code (now this test is clearer to me).

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6661

### DIFF
--- a/tests/phpunit/tests/strings-in-block/test-html.php
+++ b/tests/phpunit/tests/strings-in-block/test-html.php
@@ -58,12 +58,13 @@ class TestHTML extends \OTGS_TestCase {
 
 	/**
 	 * @test
+	 * @group wpmlcore-6661
 	 */
-	public function it_does_not_find_column_if_there_is_no_xpath() {
+	public function it_does_not_find_the_column_content_if_the_block_configuration_has_no_xpath() {
 
 		$config_option = \Mockery::mock( 'WPML_Gutenberg_Config_Option' );
 		$config_option->shouldReceive( 'get' )
-		              ->andReturn( array( 'core/column' => array( 'xpath' => array() ) ) );
+		              ->andReturn( array( 'core/column' => array() ) );
 
 		$strings_in_block = new HTML( $config_option );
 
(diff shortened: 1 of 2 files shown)


strategio:
@OnTheGoSystems/wpml review please.

kagg-design (dismissed):
+1 @strategio

tests/phpunit/tests/strings-in-block/test-html.php
```diff
@@ -58,12 +58,13 @@ public function it_finds_paragraph() {
...
 	/**
 	 * @test
+	 * @group wpmlcore-6661
 	 */
 	public function it_does_not_find_column_if_there_is_no_xpath() {
 
 		$config_option = \Mockery::mock( 'WPML_Gutenberg_Config_Option' );
 		$config_option->shouldReceive( 'get' )
-		              ->andReturn( array( 'core/column' => array( 'xpath' => array() ) ) );
+		              ->andReturn( array( 'core/column' => array() ) );
```
andreasciamanna, line 67: @strategio I would extend the test method name.

Right now, it stays "it does not find column if there is no xpath". This assumes that:
- the reader knows what a "column" is
- that we are dealing with the "column" block
- that "xpath" is in the block's configuration

I suggest you use either "it does not find the content if the block configuration has no xpath" or at least (though still incomplete) "it does not find the column block if there is no xpath".

andreasciamanna (changes requested):
@strategio all good, except for the name of the test, which I find a bit incomplete.

tests/phpunit/tests/strings-in-block/test-html.php
strategio: Renamed to `it_does_not_find_the_column_content_if_the_block_configuration_has_no_xpath`.

andreasciamanna (dismissed):

strategio:
@kagg-design, please thumb up :)

kagg-design (dismissed):
+1 @strategio

kagg-design:
@strategio sorry, again the same - forgot...

PR #51, opened 2019-06-03
Add proper support to nested lists

strategio:
The HTML lists are specific elements which requires a dedicated handling
when we register the strings or translate the block.

Summary of the changes:
1. At registration, we will exclude the nested list from the value. However, the registered string can still be composed of HTML tags.
2. When we update the block, we will also append the list child element if there's one in the original. The child element will be updated on its own.

Here's an example of nested list:

```html
<ul>
    <li>Parent <strong>1</strong>
        <ol>
            <li>Child <strong>1</strong>
                <ul>
                    <li>Grandchild 11</li>
                    <li>Grandchild 12</li>
                </ul>
            </li>
        </ol>
    </li>
    <li>Parent 2
        <ul>
            <li>Child 2</li>
        </ul>
    </li>
</ul>
```

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6613

This also requires a small change in the config file https://git.onthegosystems.com/wpml/sitepress-multilingual-cms/merge_requests/2400.

### DIFF
--- a/src/strings-in-block/class-html.php
+++ b/src/strings-in-block/class-html.php
@@ -2,7 +2,12 @@
 
 namespace WPML\PB\Gutenberg\StringsInBlock;
 
+use WPML\PB\Gutenberg\StringsInBlock\DOMHandler\StandardBlock;
+use WPML\PB\Gutenberg\StringsInBlock\DOMHandler\ListBlock;
+
 class HTML extends Base {
+
+	const LIST_BLOCK_NAME = 'core/list';
 
 	/**
 	 * @param \WP_Block_Parser_Block $block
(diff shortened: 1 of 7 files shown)


strategio:
@OnTheGoSystems/wpml, please check this one!

src/strings-in-block/class-html.php
```diff
@@ -54,22 +57,24 @@ public function update( \WP_Block_Parser_Block $block, array $string_translation
 
 		if ( $block_queries && isset( $block->innerHTML ) ) {
 
-			$dom   = $this->get_dom( $block->innerHTML );
-			$xpath = new \DOMXPath( $dom );
+			$is_list_block = $this->is_list_block( $block );
```
brucepearson: @strategio I think we're missing an abstraction here. The fact that you need to pass `$is_list_block` to several other functions indicates what we need two classes with a common interface. One for "normal" html and one for "list" html.

Could you move the code for `get_inner_HTML` and `set_element_value` into a separate class. Then at this point you instantiate the correct class depending on if it's a list block

```
if( $this->is_list_block ) {
    $block_handler = new ListBlockHandler();
} else 
   $block_handler = new StandardBlockHandler();
}

....
....
list( $text, ) = $block_handler->get_inner_HTML( $element );
....
....
$block_handler->set_element_value( $element, ..... );
```

The names I used are just for example. There are probably better names you can come up with.

brucepearson (changes requested):

src/strings-in-block/class-html.php
strategio: @brucepearson, I refactored as you suggested (which was not so quick), please check again.

brucepearson (dismissed):
@strategio Looks good. +1

strategio:
@OnTheGoSystems/wpml please one more review!

kagg-design (dismissed):
+1 @strategio

PR #52, opened 2019-06-10
Remove type hints for `\Illuminate\Support\Collection`

strategio:
This will prevent fatal errors if a new version of
`tightenco/collect` is already loaded.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6677

### DIFF
--- a/src/reusable-blocks/class-job-links.php
+++ b/src/reusable-blocks/class-job-links.php
@@ -14,7 +14,7 @@ class JobLinks {
 	/**
 	 * @param array $job_ids
 	 *
-	 * @return \Illuminate\Support\Collection
+	 * @return \Tightenco\Collect\Support\Collection|\Illuminate\Support\Collection
 	 */
 	public function get( array $job_ids ) {
 		return \collect( $job_ids )->map( function( $job_id ) {
--- a/src/reusable-blocks/class-manage-basket.php
+++ b/src/reusable-blocks/class-manage-basket.php
@@ -50,7 +50,7 @@ class ManageBasket extends Manage {
 	/**
 	 * @param array $data
 	 *
-	 * @return \Illuminate\Support\Collection
+	 * @return \Tightenco\Collect\Support\Collection|\Illuminate\Support\Collection
 	 */
 	private function extractAddedPostElements( array $data ) {
 		$source_lang  = $data['translate_from'];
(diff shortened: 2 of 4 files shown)


strategio:

src/reusable-blocks/class-job-links.php
```diff
@@ -14,7 +14,7 @@ public function __construct( \WPML_Translation_Job_Factory $job_factory ) {
 	/**
 	 * @param array $job_ids
 	 *
-	 * @return \Illuminate\Support\Collection
+	 * @return \Tightenco\Collect\Support\Collection|\Illuminate\Support\Collection
 	 */
 	public function get( array $job_ids ) {
 		return \collect( $job_ids )->map( function( $job_id ) {
```
andreasciamanna, line 20: @strategio

I think the problem is the `\` before `collect`.

The `use Illuminate\Support\Collection` should be enough to avoid any ambiguity, but if is not, why you can't just use `\Illuminate\Support\Collection\collect( $job_ids )->map( function( $job_id )` here and wherever you use this function?

andreasciamanna (changes requested):

src/reusable-blocks/class-job-links.php
andreasciamanna: Now I understand that `collect` is actually declared only in the global namespace, not with a specific one.

I still don't like this fix, but I can't even think of something better, I'm afraid.

andreasciamanna (approved):
If there isn't really a better way...

andreasciamanna:
Please wait for Bruce's feedback before merging

src/reusable-blocks/class-job-links.php
strategio: Yes @andreasciamanna, `collect` is in the global namespace.

See in the latest version https://github.com/tightenco/collect/blob/laravel-5.8-ongoing/src/Collect/Support/helpers.php#L21, it returns `\Tightenco\Collect\Support\Collection`. In the version we have in Core, it returns `\Illuminate\Support\Collection` (I can't find it on the GitHub repo).

strategio:
Closing here, was handled in https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/pull/53.

PR #55, opened 2019-06-12
Make sure that TM is using the DIC

strategio:
So more than checking if TM is active with the basket class,
we will make this check on the TM container config class to make sure
the DIC is implemented in TM (in case we are running an outdated TM version).

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlst-1925


strategio:

kagg-design (dismissed):
+1 @pierre

brucepearson:
@strategio I don't think this is needed with the fix here - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6686
TM will be blocked from loading if it's not compatible with 4.2.7

brucepearson:
@strategio I've done it a different way - https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/pull/56

PR #57, opened 2019-06-13
Prevent adding reusable blocks to the basket if the parent post has the same source lang as the target lang

strategio:
We will filter the target langs before to scan the post looking for
reusable blocks.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6689

### DIFF
--- a/src/reusable-blocks/class-manage-basket.php
+++ b/src/reusable-blocks/class-manage-basket.php
@@ -62,18 +62,40 @@ class ManageBasket extends Manage {
 		return \wpml_collect( $data['post'] )->map(
 			function( $item ) use ( $source_lang, $target_langs ) {
 				if (
-					isset( $item['checked'], $item['type'] )
-					&& 'post' === $item['type']
+					! isset( $item['checked'], $item['type'] )
+					|| 'post' !== $item['type']
 				) {
-					return new BasketElement(
-						(int) $item['checked'],
-						$source_lang,
-						$target_langs
-					);
+					return null;
 				}
 
-				return null;
+				$target_langs = $this->filterTargetLangs( $target_langs, $item['checked'] );
+
+				if ( ! $target_langs ) {
+					return null;
+				}
+
+				return new BasketElement(
+					(int) $item['checked'],
+					$source_lang,
+					$target_langs
+				);
 			}
 		)->filter();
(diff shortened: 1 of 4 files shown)


strategio:

kagg-design (dismissed):
+1 @strategio

src/reusable-blocks/class-manage-basket.php
```diff
@@ -62,18 +62,40 @@ private function extractAddedPostElements( array $data ) {
...
-					&& 'post' === $item['type']
+					! isset( $item['checked'], $item['type'] )
+					|| 'post' !== $item['type']
 				) {
-					return new BasketElement(
-						(int) $item['checked'],
-						$source_lang,
-						$target_langs
-					);
+					return null;
```
andreasciamanna, line 68: Another approach could be something like that:

```php
return \wpml_collect( $data['post'] )
	->filter( function ( $item ) {
		return isset( $item['checked'], $item['type'] ) && 'post' !== $item['type'];
	} )
	->map(
		function ( $item ) use ( $source_lang, $target_langs ) {
			$target_langs = $this->filterTargetLangs( $target_langs, $item['checked'] );
			if ( ! $target_langs ) {
				return null;
			}
			return new BasketElement(
				(int) $item['checked'],
				$source_lang,
				$target_langs
			);
		}
	)
	->filter();
```

I'm a bit confused by the purpose of `filterTargetLangs`.

I think that with some changes in the code, you could move this logic on the same filter callback (if not even an additional one) before `map` and remove the last `->filer()`.

andreasciamanna:
I think the logic is a bit overcomplicated, but it seems to do the work.

andreasciamanna (approved):

src/reusable-blocks/class-manage-basket.php
strategio: @andreasciamanna, we need to compare the `$target_langs` against each `$item['checked']` (the post ID) to remove a target lang which would be the same as the post source lang.

Even if I move some code to the first `filter` (as you did), we still need to apply `filterTargetLangs` in the `map`, and so the last `filter` is still needed.

andreasciamanna (dismissed):

PR #61, opened 2019-07-19
Do not escape `href` attribute in translated node

strategio:
I had to replace `DOMDocument::saveHTML` with `DOMDocument::saveXML`
because it's less strict. It allows to have a shortcode for the `href`
value.

As a side effect, I had to adjust some auto-closing tags in other tests.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6643

### DIFF
--- a/tests/phpunit/tests/strings-in-block/test-html.php
+++ b/tests/phpunit/tests/strings-in-block/test-html.php
@@ -146,7 +146,7 @@ class TestHTML extends \OTGS_TestCase {
 
 		$parent_1       = 'Parent <strong>1</strong>';
 		$child_11       = 'Child <strong>11</strong>';
-		$grandchild_111 = 'Grandchild<br>111';
+		$grandchild_111 = 'Grandchild<br/>111';
 		$grandchild_112 = 'Grandchild 112';
 		$child_12       = 'Child 12';
 		$parent_2       = 'Parent 2';
@@ -202,8 +202,8 @@ class TestHTML extends \OTGS_TestCase {
 		$block_name = 'core/paragraph';
 
 		$target_lang                 = 'de';
-		$original_block_inner_HTML   = 'some block content &amp; special chars<br>';
-		$translated_block_inner_HTML = 'some block content &amp; special chars ( TRANSLATED )<br>';
+		$original_block_inner_HTML   = 'some block content &amp; special chars<br/>';
+		$translated_block_inner_HTML = 'some block content &amp; special chars ( TRANSLATED )<br/>';
 
 
 		$strings = array(
@@ -366,7 +366,7 @@ class TestHTML extends \OTGS_TestCase {
 		$updated_block = $strings_in_block->update( $block, $strings, $target_lang );
 
 		$this->assertEquals(
-			'<figure class="wp-block-image"><img src="xxx" alt="' . $alt_text_translated . '" class="xxx"><figcaption>' . $caption_translated . '</figcaption></figure>',
+			'<figure class="wp-block-image"><img src="xxx" alt="' . $alt_text_translated . '" class="xxx"/><figcaption>' . $caption_translated . '</figcaption></figure>',
 			$updated_block->innerHTML
 		);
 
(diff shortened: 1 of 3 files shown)


strategio:
@OnTheGoSystems/wpml  please review!

tests/phpunit/tests/strings-in-block/test-html.php
```diff
@@ -409,6 +409,43 @@ public function it_should_update_innerContent_containing_a_slash() {
 		$this->assertEquals( '<div>' . $translated_block_inner_HTML . '</div>', $updated_block->innerHTML );
 	}
 
+	/**
+	 * @test
+	 * @group wpmlcore-6643
+	 */
+	public function it_should_update_and_NOT_escape_invalid_href() {
```
andreasciamanna, line 416: @strategio i suggest adding a similar test where you assert that the returned string is **not the same** as the escaped one.
It might seem silly to do that, but I believe it would make clearer what issue we are solving by using `saveXML` instead of `saveHTML`.

tests/phpunit/tests/strings-in-block/test-html.php
```diff
@@ -409,6 +409,43 @@ public function it_should_update_innerContent_containing_a_slash() {
...
+
+		$block               = \Mockery::mock( 'WP_Block_Parser_Block' );
+		$block->blockName    = $block_name;
+		$block->innerHTML    = '<div>' . $original_block_inner_HTML . '</div>';
+		$block->innerContent = [ $block->innerHTML ];
+
+		$updated_block = $strings_in_block->update( $block, $strings, $target_lang );
+
+		$this->assertEquals( '<div>' . $translated_block_inner_HTML . '</div>', $updated_block->innerHTML );
+	}
```
andreasciamanna, line 451: @strategio i suggest adding a similar test where you assert that the returned string is **not the same** as the escaped one.
It might seem silly to do that, but I believe it would make clearer what issue we are solving by using `saveXML` instead of `saveHTML`.

Something like that:

```php
$expected = ’<a class="tb-button__link" href="%5Btb-dynamic%20provider='__current_post'%20post='current'%20source='toolset_custom_field%7Call-fields'%20field='url'%20force-string='first'%20%5D"><span class="tb-button__content">Pincha aquí</span></a>’n

$this->assertNotSame( $expected, $updated_block->innerHTML );
```

That would work more as a documentation test than an actual test.

tests/phpunit/tests/strings-in-block/test-html.php
```diff
@@ -409,6 +409,43 @@ public function it_should_update_innerContent_containing_a_slash() {
...
+		);
+
+		$block               = \Mockery::mock( 'WP_Block_Parser_Block' );
+		$block->blockName    = $block_name;
+		$block->innerHTML    = '<div>' . $original_block_inner_HTML . '</div>';
+		$block->innerContent = [ $block->innerHTML ];
+
+		$updated_block = $strings_in_block->update( $block, $strings, $target_lang );
+
+		$this->assertEquals( '<div>' . $translated_block_inner_HTML . '</div>', $updated_block->innerHTML );
```
andreasciamanna, line 446: I think you can use `assertAttributeSame`/`assertAttributeSame` to  test the attribute of a DOM eLement.

andreasciamanna: Tough, with the wrapper `<div>`, it might be problematic.

andreasciamanna (changes requested):

tests/phpunit/tests/strings-in-block/test-html.php
strategio: Done @andreasciamanna, indeed it's useful for documentation.

tests/phpunit/tests/strings-in-block/test-html.php
strategio: I never used this kind of assertion, I'll probably use it in the future. Here, I prefer to keep testing the string because this is what we get as the output.

andreasciamanna (dismissed):

strategio:
@OnTheGoSystems/wpml one more review please.

kagg-design (dismissed):
+1 @pierre

PR #65, opened 2019-09-04
Handle link type fields for gutenberg blocks

brucepearson:
Handle link type fields for gutenberg blocks - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6471

### DIFF
--- a/src/class-wpml-gutenberg-strings-registration.php
+++ b/src/class-wpml-gutenberg-strings-registration.php
@@ -20,16 +20,26 @@ class WPML_Gutenberg_Strings_Registration {
 	/** @var array $leftover_strings */
 	private $leftover_strings;
 
+	/** @var WPML_Translate_Link_Targets $translate_link_targets */
+	private $translate_link_targets;
+
+	/** @var callable $set_link_translations */
+	private $set_link_translations;
+
 	public function __construct(
 		WPML\PB\Gutenberg\StringsInBlock\StringsInBlock $strings_in_blocks,
 		WPML_ST_String_Factory $string_factory,
 		WPML_PB_Reuse_Translations $reuse_translations,
-		WPML_PB_String_Translation $string_translation
+		WPML_PB_String_Translation $string_translation,
+		WPML_Translate_Link_Targets $translate_link_targets,
+		callable $set_link_translations
 	) {
-		$this->strings_in_blocks = $strings_in_blocks;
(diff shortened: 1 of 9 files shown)


tests/phpunit/tests/test-wpml-gutenberg-string-registration.php
```diff
@@ -426,3 +502,18 @@ private function getStringsInBlock( $config_option ) {
 		return new WPML\PB\Gutenberg\StringsInBlock\Collection( $string_parsers );
 	}
 }
+
+class ClosureMock {
```
brucepearson, line 571: **Note:** I couldn't find a standard way of mocking a closure so I created this class to help. I plan to move this to a separate test library when I introduce the Functional library into WPML.

src/class-wpml-gutenberg-strings-registration.php
```diff
@@ -76,6 +90,8 @@ private function register_blocks( array $blocks, array $package_data ) {
 					$string->type
 				);
 
+				$string_id = apply_filters( 'wpml_string_id_from_package', 0, $package_data, $string->id, $string->value );
```
strategio: @brucepearson, I think you can move this line to just before 106.

src/strings-in-block/class-html.php
```diff
@@ -63,6 +65,7 @@ public function update( \WP_Block_Parser_Block $block, array $string_translation
 			$xpath      = new \DOMXPath( $dom );
 
 			foreach ( $block_queries as $query ) {
+				list( $query, $definedType ) = XPath::parse( $query );
```
strategio: I think `$definedType` is not used here... maybe `list( $query )` only ?

strategio (changes requested):
I just a couple of small suggestion @brucepearson.

src/class-wpml-gutenberg-strings-registration.php
brucepearson: @strategio Done

src/strings-in-block/class-html.php
brucepearson: @strategio Done

strategio (dismissed):
+1 @brucepearson

strategio (dismissed):

gordonsh24 (dismissed):
@brucepearson +1

PR #66, opened 2019-09-24
Use correct options for encoding JSON attributes.

dgwatkins:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6834

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,20 +14,20 @@
       "src/"
     ]
   },
-  "repositories": [
-    {
+  "repositories": {
+    "collect": {
       "type": "vcs",
       "url": "https://github.com/OnTheGoSystems/collect.git"
     }
-  ],
+  },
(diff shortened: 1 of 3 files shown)


andreasciamanna:
@OnTheGoSystems/compatibility

strategio (dismissed):
@dgwatkins, I mentionned in the ticket that we should try to have a CC test for this issue. That's because it involves the way data is sent to the server and the implementation of WP and WPML Core.

dgwatkins:
Sorry @strategio 

I just dismissed your review because I saw how Konrad tested this sort of change and did the same.

As for the CC test,  I will make sure its done.

brucepearson (dismissed):
@dgwatkins It looks good to me. +1

strategio (dismissed):
+1

PR #67, opened 2019-11-12
Only output self closing tags supported in HTML5

brucepearson:
Only output self closing tags supported in HTML5 - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6936

### DIFF
--- a/src/strings-in-block/dom-handler/dom-handle.php
+++ b/src/strings-in-block/dom-handler/dom-handle.php
@@ -117,4 +117,25 @@ abstract class DOMHandle {
 	private function cloneNodeWithoutChildren( \DOMNode $element ) {
 		return $element->cloneNode( false );
 	}
+
+	protected function getAsHTML5( \DOMNode $element ) {
+		return strtr( $element->ownerDocument->saveXML( $element, LIBXML_NOEMPTYTAG ),
+			[
+				'></area>'   => '/>',
+				'></base>'   => '/>',
+				'></br>'     => '/>',
+				'></col>'    => '/>',
+				'></embed>'  => '/>',
+				'></hr>'     => '/>',
+				'></img>'    => '/>',
+				'></input>'  => '/>',
+				'></link>'   => '/>',
+				'></meta>'   => '/>',
+				'></param>'  => '/>',
+				'></source>' => '/>',
+				'></track>'  => '/>',
+				'></wbr>'    => '/>',
+			] );
+	}
+
 }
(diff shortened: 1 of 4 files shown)


kagg-design (dismissed):
+1 @brucepearson

src/strings-in-block/dom-handler/dom-handle.php
```diff
@@ -117,4 +117,25 @@ public function setElementValue( \DOMNode $element, $value ) {
 	private function cloneNodeWithoutChildren( \DOMNode $element ) {
 		return $element->cloneNode( false );
 	}
+
+	protected function getDOMString( \DOMNode $element ) {
```
andreasciamanna: @BruceOnTheGo this is a class called `DOMHandle` under the `DOMHandler` namespace: I think including "DOM" in the name of the method is a bit redundant.

Instead, since this method doesn't simply return a string form of the DOM Node, it's more important to explain we are returning an HTML (HTML5?) string.

This information can be added as part of the method name (anything like `getHTML`, `getHTML5`, `toHTML`, `toHTML5`, `getHTMLString`, etc.) or as a description in the PHPDoc.

tests/phpunit/tests/strings-in-block/test-html.php
```diff
@@ -361,7 +361,7 @@ public function it_updates_image() {
 
 		$block            = \Mockery::mock( 'WP_Block_Parser_Block' );
 		$block->blockName = $block_name;
-		$block->innerHTML = '<figure class="wp-block-image"><img src="xxx" alt="' . $alt_text . '" class="xxx" /><figcaption>' . $caption . '</figcaption></figure>';
+		$block->innerHTML = '<figure class="wp-block-image"><img src="xxx" alt="' . $alt_text . '" class="xxx"/><figcaption>' . $caption . '</figcaption></figure>';
```
andreasciamanna: @BruceOnTheGo shouldn't the code work with or without space?

andreasciamanna (dismissed):
@BruceOnTheGo overall, the fix is fine, but I have some trouble understanding the small edit you've made in the existing test, and a comment about the naming of the new `getDOMString` method.

tests/phpunit/tests/strings-in-block/test-html.php
brucepearson: I removed the space during one of the iterations of the fix. It works with or without the space.

PR #68, opened 2019-11-13
Apply a filter when we are updating a translation for a Gutenberg block

brucepearson:
Apply a filter when we are updating a translation for a Gutenberg block - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6941

### DIFF
--- a/src/strings-in-block/class-html.php
+++ b/src/strings-in-block/class-html.php
@@ -80,7 +80,9 @@ class HTML extends Base {
 					}
 				}
 			}
-			list( $block->innerHTML, ) = $dom_handle->getFullInnerHTML( $dom->documentElement );
+			list( $translation, ) = $dom_handle->getFullInnerHTML( $dom->documentElement );
+			$block = apply_filters( 'wpml_updating_block_translation', $block, $translation );
+			$block->innerHTML = $translation;
 
 		} else {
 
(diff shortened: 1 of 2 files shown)


src/strings-in-block/class-html.php
```diff
@@ -89,7 +91,9 @@ public function update( \WP_Block_Parser_Block $block, array $string_translation
 				isset( $string_translations[ $string_id ][ $lang ] ) &&
 				ICL_TM_COMPLETE == $string_translations[ $string_id ][ $lang ]['status']
 			) {
-				$block->innerHTML = $string_translations[ $string_id ][ $lang ]['value'];
+				$translation = $string_translations[ $string_id ][ $lang ]['value'];
+				$block = apply_filters( 'wpml_updating_block_translation', $block, $translation );
+				$block->innerHTML = $translation;
```
andreasciamanna, line 96: @brucepearson I think you can extract the two lines into a private method.

Something like that:

```php
private function filter_block_translation( $block, $translation ) {
	$block            = apply_filters( 'wpml_updating_block_translation', $block, $translation );
	$block->innerHTML = $translation;

	return $block;
}
```

Then here an above you can call `$block = $this->filter_block_translation( $block, $translation );`.

(or even pass `$block` by reference and avoid to return it)

andreasciamanna (approved):
It looks fine to me. I've only added a comment to prevent duplication (especially because of the duplicated reference to the same filter).

strategio (approved):
@brucepearson, I agree with Andrea, it's better if we can avoid the duplication. Please also add a PHPDoc to this new filter hook.

brucepearson:
Closing this as there's already a filter that can be used.

PR #69, opened 2019-11-14
Toolset Blocks - Fields and Text translation is missing p and br tags.

kagg-design:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6938

### DIFF
--- a/src/class-wpml-gutenberg-integration.php
+++ b/src/class-wpml-gutenberg-integration.php
@@ -5,10 +5,11 @@
  */
 class WPML_Gutenberg_Integration implements \WPML\PB\Gutenberg\Integration {
 
-	const PACKAGE_ID              = 'Gutenberg';
-	const GUTENBERG_OPENING_START = '<!-- wp:';
-	const GUTENBERG_CLOSING_START = '<!-- /wp:';
-	const CLASSIC_BLOCK_NAME      = 'core/classic-block';
+	const PACKAGE_ID                         = 'Gutenberg';
+	const GUTENBERG_OPENING_START            = '<!-- wp:';
+	const GUTENBERG_CLOSING_START            = '<!-- /wp:';
+	const CLASSIC_BLOCK_NAME                 = 'core/classic-block';
+	const TOOLSET_FIELDS_AND_TEXT_BLOCK_NAME = 'toolset-blocks/fields-and-text';
 
 	/**
 	 * @var WPML\PB\Gutenberg\StringsInBlock\StringsInBlock
@@ -194,6 +195,10 @@ class WPML_Gutenberg_Integration implements \WPML\PB\Gutenberg\Integration {
 
 		} else {
 			$content = $block->innerHTML;
+		}
+
+		if ( self::TOOLSET_FIELDS_AND_TEXT_BLOCK_NAME === $block->blockName ) {
+			$content = wpautop( $content );
 		}
 
 		return $content;
--- a/src/class-wpml-gutenberg-integration.php
+++ b/src/class-wpml-gutenberg-integration.php
@@ -5,11 +5,10 @@
  */
 class WPML_Gutenberg_Integration implements \WPML\PB\Gutenberg\Integration {
 
-	const PACKAGE_ID                         = 'Gutenberg';
-	const GUTENBERG_OPENING_START            = '<!-- wp:';
-	const GUTENBERG_CLOSING_START            = '<!-- /wp:';
-	const CLASSIC_BLOCK_NAME                 = 'core/classic-block';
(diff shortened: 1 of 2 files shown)


kagg-design:
@wpml-code-review @brucepearson Please review.

strategio (commented):
@kagg-design, if this needs to be fixed urgently, I am fine with this as a temporary fix.

However, we should move this to somewhere else: A compatibility class for Toolset or another PB package dedicated to Toolset (I think the compatibility class should be enough).

This `\WPML_Gutenberg_Integration` is already too big and we should not introduce specific case anymore (it should go through a filter).

strategio (dismissed):
As discussed in MM, I'll give my approval if this is urgent. In that case, @kagg-design please create another ticket so we can re-factor the code later.

src/class-wpml-gutenberg-integration.php
```diff
@@ -196,6 +197,10 @@ private function render_inner_HTML( $block ) {
 			$content = $block->innerHTML;
 		}
 
+		if ( self::TOOLSET_FIELDS_AND_TEXT_BLOCK_NAME === $block->blockName ) {
```
brucepearson: @kagg-design Do you know why we need this only for this block and not for other blocks? eg. What is between this and the wp/paragraph block?
Is there a way to make this more general so it can be used for other blocks if they need it also?

kagg-design: @brucepearson I have tested core/paragraph block and it is properly prepared (has `p` and `br` at this point).

andreasciamanna: As it seems this behaviour is specific to Toolset, or at least, not the standard behaviour in Gutenberg, I suggest we allow third-party plugins to control that behaviour.

# Option 1

Replace `self::TOOLSET_FIELDS_AND_TEXT_BLOCK_NAME === $block->blockName` with `apply_filters( 'wpml_gutenberg_inner_HTML_content_autop,  false, $block, $content )`.

# Option 2

Provide a more generic filter such as `apply_filters( 'wpml_gutenberg_inner_HTML_content,  $content, $block )`.

# Option 3

Same as Option 2, but add the `wpml_inner_HTML_content` filter to the main **wpml-page-builders** project and include a third argument which specifies the type of integration (Gutenberg, Elementor, Cornerstone, etc.).

kagg-design: @andreasciamanna Unfortunately, there is a problem with filtering. Toolset Views simply do not init class `\ToolsetBlocks\Block\Style\Block\FieldsAndText`. This class is the only logical point to add a filter. 

I have added filter to `wp-content/plugins/views/vendor/toolset/blocks/server/Block/Style/Block/Factory.php` and it works, but looks like a hack: https://git.onthegosystems.com/toolset/toolset-blocks/merge_requests/193/diffs

kagg-design: More brave approach is to make `wpautop` without filter, like we do in `\WPML_Gutenberg_Integration::render_block` for Classic Block.

With those blocks I tested (for instance, simple paragraph), nothing wrong happens. Block inner html comes as `\n<p>text1<br />\ntext2<p>\n` and it is OK. After `wpautop` only first `\n` is missed, so it becomes `<p>text1<br />\ntext2<p>\n`.

andreasciamanna: @kagg-design Mine was meant to be food for thought: let third-party plugins hook into our code.

Then I suggested three possible options.

I wasn't really suggesting to do any change at this stage since the plan was to let Toolset fix the problem for now: did we do that?

src/class-wpml-gutenberg-integration.php
```diff
@@ -193,7 +193,7 @@ private function render_inner_HTML( $block ) {
 			}
 
 		} else {
-			$content = $block->innerHTML;
+			$content = apply_filters( 'wpml_gutenberg_render_inner_HTML', $block->innerHTML, $block );
```
brucepearson, line 196: @kagg-design There's already an existing filter that can be used instead of adding a new one. `wpml_update_strings_in_block`

src/class-wpml-gutenberg-integration.php
brucepearson: The best approach for this is to add a format attribute to the config file and then handle in our code according to that. **format="wpautop"**

```
<gutenberg-block type="toolset-blocks/fields-and-text" translate="1">
	<xpath format="wpautop">//div[@class='tb-fields-and-text']</xpath>
</gutenberg-block>
```

brucepearson:
I'm closing this as we should use the existing filter rather than adding a new one.

PR #73, opened 2019-12-03
Do not escape forward slash (`/`) in block attributes

strategio:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6995

### DIFF
--- a/tests/phpunit/tests/test-wpml-gutenberg-integration.php
+++ b/tests/phpunit/tests/test-wpml-gutenberg-integration.php
@@ -144,6 +144,7 @@ class Test_WPML_Gutenberg_Integration extends OTGS_TestCase {
 	/**
 	 * @test
 	 * @group wpmlcore-6221
+	 * @group wpmlcore-6995
 	 */
 	public function it_updates_translated_page() {
 		$original_post               = \Mockery::mock( 'WP_Post' );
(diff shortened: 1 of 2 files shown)


strategio:

tests/phpunit/tests/test-wpml-gutenberg-integration.php
```diff
@@ -160,6 +161,7 @@ public function it_updates_translated_page() {
 			'att_2' => 'value_2',
 			'att_3' => 'polish żółć',
 			'att_4' => 'Value with "quotes"',
+			'att_5' => '<div>[my_shortcode][/my_shortcode]</div>',
```
andreasciamanna: @strategio don't you think we need a case where the shortcode has attributes and actual content in it?

andreasciamanna (changes requested):

tests/phpunit/tests/test-wpml-gutenberg-integration.php
strategio: @andreasciamanna, I started creating an independent test method for this case and it was very big (all tests are pretty long in this class because it's doing a lot of things). Then I saw this test where we are testing different attributes and their encoding. I think this is more self-explainable than an independent method and we keep all attribute cases at the same place.

andreasciamanna: @strategioI wasn't suggesting to create a separate test.

This test has a shortcode with a closing tag, which address your fix.
However, a shortcode with a closing tag may have attributes and even more likely, content. Like `[shortcode a=”1” b=”2”]some content<br/>with html[/shortcode]`.

Even if you don't test the actual handling of the attributes and the content, I think we must ensure your fix still work with this type of shortcode.

strategio:
Andrea, I will improve the test later, I needs to go out for an appointment
now.

andreasciamanna (dismissed):
Assuming the test will be improved (again, I'm not asking for a separate test), and to expedite the review, I'm fine with it.

andreasciamanna (dismissed):

andreasciamanna:
@OnTheGoSystems/wpml please review this

gordonsh24 (dismissed):

PR #74, opened 2019-12-04
Introduce classes to convert Ids in block

strategio:
**Is required for https://git.onthegosystems.com/wpml/woocommerce-multilingual/merge_requests/1179**

- `BlockAttributes`
- `TagAttributes`
- `NullConverter` (do nothing)
- `Composite` (to composer converters)

This is required to convert blocks in WooCommerce but it can also be
used by other plugins.

https://onthegosystems.myjetbrains.com/youtrack/issue/wcml-2676

### DIFF
--- a/src/IdsInBlock/Base.php
+++ b/src/IdsInBlock/Base.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace WPML\PB\Gutenberg\ConvertIdsInBlock;
+
+class Base {
+
+	/**
+	 * @param array $block
+	 *
+	 * @return array
+	 */
+	public function convert( array $block ) {
+		return $block;
+	}
+
+	/**
+	 * @param array|int $ids
+	 * @param string    $elementType
+	 *
+	 * @return array|int
+	 */
+	public static function convertIds( $ids, $elementType ) {
+		$getTranslation = function( $id ) use ( $elementType ) {
+			return (int) wpml_object_id_filter( $id, $elementType );
+		};
+
+		if ( is_array( $ids ) ) {
+			return wpml_collect( $ids )->map( $getTranslation )->toArray();
+		}
+
+		return $getTranslation( $ids );
+	}
+}
--- a/src/IdsInBlock/BlockAttributes.php
+++ b/src/IdsInBlock/BlockAttributes.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace WPML\PB\Gutenberg\ConvertIdsInBlock;
+
+class BlockAttributes extends Base {
+
+	private $attributesToConvert;
+
+	public function __construct( array $attributesToConvert ) {
+		$this->attributesToConvert = $attributesToConvert;
+	}
+
+	public function convert( array $block ) {
+		foreach ( $this->attributesToConvert as $attributeConfig ) {
+			$name = $attributeConfig['name'];
+
+			if ( isset( $block['attrs'][ $name ] ) ) {
+				$block['attrs'][ $name ] = self::convertIds(
+					$block['attrs'][ $name ],
+					$attributeConfig['type']
+				);
+			}
+		}
+
+		return $block;
+	}
+}
--- a/src/strings-in-block/dom-handler/dom-handle.php
+++ b/src/strings-in-block/dom-handler/dom-handle.php
@@ -29,12 +29,8 @@ abstract class DOMHandle {
 		$dom = new \DOMDocument();
 		\libxml_use_internal_errors( true );
 		$html = mb_convert_encoding( $html, 'HTML-ENTITIES', 'UTF-8' );
-		$dom->loadHTML( '<div>' . $html . '</div>' );
+		$dom->loadHTML( '<div>' . $html . '</div>', LIBXML_HTML_NOIMPLIED | LIBXML_HTML_NODEFDTD );
 		\libxml_clear_errors();
-
-		// Remove doc type and <html> <body> wrappers
-		$dom->removeChild( $dom->doctype );
-		$dom->replaceChild( $dom->firstChild->firstChild->firstChild, $dom->firstChild );
 
 		return $dom;
 	}
--- a/src/strings-in-block/dom-handler/dom-handle.php
+++ b/src/strings-in-block/dom-handler/dom-handle.php
@@ -121,7 +121,8 @@ abstract class DOMHandle {
 	}
 
 	protected function getAsHTML5( \DOMNode $element ) {
-		return strtr( $element->ownerDocument->saveXML( $element, LIBXML_NOEMPTYTAG ),
+		return strtr(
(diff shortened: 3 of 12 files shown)


src/strings-in-block/dom-handler/dom-handle.php
```diff
@@ -121,7 +121,13 @@ private function cloneNodeWithoutChildren( \DOMNode $element ) {
 	}
 
 	protected function getAsHTML5( \DOMNode $element ) {
-		return strtr( $element->ownerDocument->saveXML( $element, LIBXML_NOEMPTYTAG ),
+		$html = str_replace(
```
strategio: I think this is due to the recent changes. When we use `saveXML`, it wraps everything inside `<body>`. So I had to remove it.

brucepearson: @strategio Are you sure about this. Wouldn't everything else be broken if this code was wrapping in a body tag.

src/IdsInBlock/Base.php
```diff
@@ -0,0 +1,31 @@
...
+	public static function convertIds( $ids, $elementType ) {
+		$getTranslation = function( $id ) use ( $elementType ) {
+			return (int) wpml_object_id_filter( $id, $elementType );
+		};
+
+		if ( is_array( $ids ) ) {
+			return wpml_collect( $ids )->map( $getTranslation )->toArray();
+		} else {
+			return wpml_collect( [ $ids ] )->map( $getTranslation )->first();
+		}
```
andreasciamanna, line 29: @strategio 

If you don't have an array, you convert the value to an array, pass it to `wpml_collect`, then `map` (for one item) and then you get the first element.

That looks a bit overkill to me.

Wouldn't this achieve the same result?

```php
if ( is_array( $ids ) ) {
	return wpml_collect( $ids )->map( $getTranslation )->toArray();
}

return $getTranslation( $ids );
```

src/IdsInBlock/BlockAttributes.php
```diff
@@ -0,0 +1,26 @@
...
+	public function convert( array $block ) {
+		foreach ( $this->attributesToConvert as $attributeConfig ) {
+
+			if ( isset( $block['attrs'][ $attributeConfig['name'] ] ) ) {
+				$block['attrs'][ $attributeConfig['name'] ] = self::convertIds(
+					$block['attrs'][ $attributeConfig['name'] ],
+					$attributeConfig['type']
+				);
+			}
+		}
```
andreasciamanna, line 23: @strategio that's a bit hard to read. Can you extract some of the code?

Something like that:

```php
$attrs = $block['attrs'];

foreach ( $this->attributesToConvert as $attributeConfig ) {
	$name = $attributeConfig['name'];
	if ( isset( $attrs[ $name ] ) ) {
		$attrs[ $name ] = self::convertIds(
			$attrs[ $name ],
			$attributeConfig['type']
		);
	}
}

$block['attrs'] = $attrs;
```

src/IdsInBlock/NullConvert.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace WPML\PB\Gutenberg\ConvertIdsInBlock;
+
+class NullConvert extends Base{
```
andreasciamanna: @strategio That "Null" looks a bit misleading to me. What about `NoConvert`?

As a matter of fact, I don't think you need this extra class.

The `convert` method in `\WPML\PB\Gutenberg\ConvertIdsInBlock\Base` can be changed into a normal method with the same body as you have here.
Also, `\WPML\PB\Gutenberg\ConvertIdsInBlock\Base`  won't need to be declared as an abstract class anymore.

With this change, all other classes, instead of just "implementing" the abstract method, will "override" it (though I think that, at least in theory, each child class is supposed to call `parent::convert( $block);` before overriding it).

In WCML, the only change you would need is to replace ´NullConvert´ with `Base`.

src/strings-in-block/dom-handler/dom-handle.php
andreasciamanna: @BruceOnTheGo @strategio 

I think the problem is when we load the HTML (I assume it's in `\WPML\PB\Gutenberg\StringsInBlock\DOMHandler\DOMHandle::getDom`).

I did a quick test with this snippet:

```php
$html = <<<'HTML'
<form action="http://maps.google.com/maps" method="get" target="_blank">
   <label for="saddr">Enter your location</label>
   <input type="text" name="saddr" />
   <input type="hidden" name="daddr" value="350 5th Ave New York, NY 10018 (Empire State Building)" />
   <input type="submit" value="Get directions" />
</form>
HTML;


$dom = new \DOMDocument();
$dom->loadHTML( $html );
//$dom->loadHTML( $html, LIBXML_HTML_NOIMPLIED );

$children = $dom->childNodes;

foreach ( $children as $child ) {
	echo $child->ownerDocument->saveXML( $child, LIBXML_NOEMPTYTAG );
}
```

I doesn't only add the `<body>` tag. In my case, I also get `<html>`.

Now, if you comment out `$dom->loadHTML( $html );` and uncomment `$dom->loadHTML( $html, LIBXML_HTML_NOIMPLIED );`, no extra tags are added.

In both cases, the string always begins with `<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN" "http://www.w3.org/TR/REC-html40/loose.dtd">`.

`$dom->loadHTML( $html, LIBXML_HTML_NOIMPLIED | LIBXML_HTML_NODEFDTD );` will remove the DTD as well.

It's all documented here, by the way: https://www.php.net/manual/en/libxml.constants.php

andreasciamanna (changes requested):

src/strings-in-block/dom-handler/dom-handle.php
brucepearson: We had some magic in `DOMHandle::getDom` to handle this. But with the suggestion from @andreasciamanna it can probably be simplified.

src/IdsInBlock/Base.php
strategio: Done!

src/IdsInBlock/BlockAttributes.php
strategio: Done

src/IdsInBlock/NullConvert.php
strategio: Done

src/strings-in-block/dom-handler/dom-handle.php
strategio: Initially I was not using  `DOMHandle::getDom` in my new code. Now this is done and I don't need the change here anymore.

I applied Andrea' suggestion and it seems to do the trick.

strategio:
@andreasciamanna, I think I addressed all your comments.

andreasciamanna (dismissed):
I'm still off, but I didn't want to get this pr stuck.

It looks good now!

PR #76, opened 2020-01-08
Improve render_block method for self-closing blocks

strategio:
FTR, WP 5.3.1 comes with a new function `get_comment_delimited_block_content` but it has several glitches and I am not sure it covers all the little fixes that we added recently. Maybe we can replace our code with this in the future, but not now.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7038


brucepearson (approved):
@strategio Looks good. +1

kagg-design (approved):
+1 @strategio

PR #77, opened 2020-01-17
Convert and fallback to original if element is displayed as translated

strategio:
If the ID conversion returns a falsy value and the element is displayed
as translated, we will use the original ID.

https://onthegosystems.myjetbrains.com/youtrack/issue/wcml-3082


strategio:

brucepearson (approved):
@strategio Looks good. +1

ghost (approved):
+1

PR #78, opened 2020-05-05
Add support for `\DOMText` nodes

strategio:
These nodes are provided when we have XPaths of that kind `/div/text()` and for this kind of block content `<div>Some text<span>Not to translate</span></div>`.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7234


brucepearson (approved):
@strategio Looks good. +1

dgwatkins (approved):
:+1:

PR #79, opened 2020-05-13
Set the string type to `AREA` when it's longer than 80 chars

strategio:
The length 80 is an arbitrary choice from me, but seems a good
compromise.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-6351


dgwatkins (approved):
Good one :+1:

brucepearson (approved):

PR #80, opened 2020-05-22
Handle Gutenberg Block defined as LINK type in wpml-config.xml

GeorgeBotsev:
When the link is external, it is set by other logic to Line
However, if the link is internal, it shoudl be translated by WPML and
it should be set to LINK type

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7069

### DIFF
--- a/src/strings-in-block/class-base.php
+++ b/src/strings-in-block/class-base.php
@@ -87,7 +87,7 @@ abstract class Base implements StringsInBlock {
 		if ( strpos( $string, '<' ) !== false ) {
 			$type = 'VISUAL';
 		}
-		if ( strpos( $string, '://' ) !== false ) {
+		if ( filter_var( $string, FILTER_VALIDATE_URL ) ) {
 			$type = 'LINK';
 		}
 
--- a/src/strings-in-block/class-base.php
+++ b/src/strings-in-block/class-base.php
@@ -87,6 +87,9 @@ abstract class Base implements StringsInBlock {
 		if ( strpos( $string, '<' ) !== false ) {
 			$type = 'VISUAL';
 		}
+		if ( strpos( $string, '://' ) !== false ) {
+			$type = 'LINK';
+		}
 
 		return $type;
 	}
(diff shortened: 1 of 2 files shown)


src/strings-in-block/class-base.php
```diff
@@ -87,6 +87,9 @@ public static function get_string_type( $string ) {
 		if ( strpos( $string, '<' ) !== false ) {
 			$type = 'VISUAL';
 		}
+		if ( strpos( $string, '://' ) !== false ) {
```
dgwatkins: @GeorgeBotsev 
The check looks a bit greedy ... for example, if the string has a paragraph with a link it would be matched.

kkarpieszuk: @GeorgeBotsev this does not guarantee this will handle only links. Should this mark strings like "Page addresses always have :// after the protocol name" as LINK too?

I suggest:
```
if (filter_var($string, FILTER_VALIDATE_URL) !== false)
```

kkarpieszuk (commented):
George, I suggested a different approach. Plus like David suggested, please add phpunit tests (I assume this is the hard part for you if you've never done, so ping me if you need help)

dgwatkins (dismissed):
Im resolving this thread because @kkarpieszuk suggestion is good ... 
only that it should be a Yoda condition.

src/strings-in-block/class-base.php
kkarpieszuk: George, please fix this and do another `git commit -a -m '[fixup] better url validation'`

kkarpieszuk (approved):

kkarpieszuk:
+1

GeorgeBotsev:
I am still checking some cases related to this right now...

dgwatkins (approved):
:+1: @GeorgeBotsev

strategio:
@GeorgeBotsev, with https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/pull/82 I believe now that this PR can be closed. Let me know your thoughts.

strategio:
Closing here, replaced by https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/pull/80.

PR #82, opened 2020-05-26
Fix xpath config parsing when a block has only 1 element and contains a type attribute

strategio:
The array produced from the XML config does not always have the same
shape if it has a single element or several ones.

1. single element:

```php
$xpath = [
    'value' => '//my/xpath',
    'attr'  => [ 'type' => 'link' ],
];
```

2. multiple elements:

```php
$xpath = [
    [
        'value' => '//my/xpath1',
        'attr'  => [ 'type' => 'link' ],
    ],
    [
        'value' => '//my/xpath2',
        'attr'  => [ 'type' => 'link' ],
    ],
];
```

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7069

**Important note:** I am trying to understand what has changed in the previous PR (https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/pull/65/files#diff-48a751ae78075d37ac563a6f4ca23b4bR47) because we were still applying the data normalization at the same place... I suspect that the data shape changed when we added the `type` attribute to the config `xsd` file (se [here](https://git.onthegosystems.com/wpml/sitepress-multilingual-cms/-/merge_requests/2489/diffs#c30cd42a1989b53a111f60442d1317b76ce3ae7e_278_277)).

### DIFF
--- a/src/class-wpml-gutenberg-config-option.php
+++ b/src/class-wpml-gutenberg-config-option.php
@@ -43,7 +43,7 @@ class WPML_Gutenberg_Config_Option {
 			$block_name                     = $block_config['attr']['type'];
 			$blocks[ $block_name ]['xpath'] = array();
 
-			foreach ( $block_config['xpath'] as $xpaths ) {
+			foreach ( $this->normalize_key_data( $block_config['xpath'] ) as $xpaths ) {
 				$xpaths                         = XPath::normalize( $xpaths );
 				$blocks[ $block_name ]['xpath'] = array_merge( $blocks[ $block_name ]['xpath'], array_values( $xpaths ) );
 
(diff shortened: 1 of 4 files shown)


dgwatkins (approved):
:+1: good one

src/class-wpml-gutenberg-config-option.php
```diff
@@ -43,7 +43,7 @@ private function add_block_xpaths( array $blocks, array $block_config ) {
 			$block_name                     = $block_config['attr']['type'];
 			$blocks[ $block_name ]['xpath'] = array();
 
-			foreach ( $block_config['xpath'] as $xpaths ) {
+			foreach ( $this->normalize_key_data( $block_config['xpath'] ) as $xpaths ) {
```
brucepearson, line 46: @strategio This seems correct to me and should fix the problem at the source.

brucepearson (approved):
@strategio Looks good. +1

PR #83, opened 2020-06-08
Allow shortcodes to be processed along with gutenberg blocks - https:…

brucepearson:
…//onthegosystems.myjetbrains.com/youtrack/issue/wpmltm-3823

### DIFF
--- a/src/strings-in-block/class-html.php
+++ b/src/strings-in-block/class-html.php
@@ -2,6 +2,7 @@
 
 namespace WPML\PB\Gutenberg\StringsInBlock;
 
+use WPML\FP\Obj;
 use WPML\PB\Gutenberg\StringsInBlock\DOMHandler\StandardBlock;
 use WPML\PB\Gutenberg\StringsInBlock\DOMHandler\ListBlock;
 use WPML\PB\Gutenberg\XPath;
@@ -69,15 +70,14 @@ class HTML extends Base {
 				$elements = $xpath->query( $query );
 				foreach ( $elements as $element ) {
 					list( $text, ) = $dom_handle->getPartialInnerHTML( $element );
-					$string_id = $this->get_string_id( $block->blockName, $text );
-					if (
-						isset( $string_translations[ $string_id ][ $lang ] ) &&
-						ICL_TM_COMPLETE == $string_translations[ $string_id ][ $lang ]['status']
-					) {
-						$translation = $string_translations[ $string_id ][ $lang ]['value'];
-						$block       = self::update_string_in_innerContent( $block, $element, $translation );
-						$dom_handle->setElementValue( $element, $translation );
-					}
+					$block = $this->updateTranslationInBlock(
+						$text,
+						$lang,
+						$block,
+						$string_translations,
+						$element,
+						$dom_handle
+					);
 				}
 			}
 			list( $block->innerHTML, ) = $dom_handle->getFullInnerHTML( $dom->documentElement );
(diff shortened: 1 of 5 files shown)


src/strings-in-block/class-html.php
```diff
@@ -174,4 +174,34 @@ private function get_dom_handler( \WP_Block_Parser_Block $block ) {
...
+		$translation                = null;
+		$translationFromPageBuilder = apply_filters( 'wpml_pb_update_translations_in_content', $text, $lang );
+		if ( $translationFromPageBuilder === $text ) {
+			$string_id = $this->get_string_id( $block->blockName, $text );
+			if ( Obj::path( [ $string_id, $lang, 'status'], $string_translations ) == ICL_TM_COMPLETE ) {
+				$translation = $string_translations[ $string_id ][ $lang ]['value'];
+			}
+		} else {
+			$translation = $translationFromPageBuilder;
+		}
```
andreasciamanna: @BruceOnTheGo since both branches of the conditional set a value for $translation, it might be a good idea to extra the code into a `getTranslationFromPageBuilder` private method.

Something like that:

```php
private function updateTranslationInBlock( $text, $lang, \WP_Block_Parser_Block $block, array $string_translations, $element, $dom_handle ) {
	$translation = $this->getTranslationFromPageBuilder($text, $lang,$block,$string_translations);
	if ( $translation ) {
		$block = self::update_string_in_innerContent( $block, $element, $translation );
		$dom_handle->setElementValue( $element, $translation );
	}

	return $block;
}

private function getTranslationFromPageBuilder(  $text, $lang, \WP_Block_Parser_Block $block, array $string_translations ) {
	$translation                = null;
	$translationFromPageBuilder = apply_filters( 'wpml_pb_update_translations_in_content', $text, $lang );
	if ( $translationFromPageBuilder === $text ) {
		$string_id = $this->get_string_id( $block->blockName, $text );
		if ( Obj::path( [ $string_id, $lang, 'status' ], $string_translations ) == ICL_TM_COMPLETE ) {
			return $string_translations[ $string_id ][ $lang ]['value'];
		}
	}

	return $translationFromPageBuilder;
}
```

andreasciamanna (dismissed):
It looks alright.

Just added a comment for what I think would improve the readability of the code.

src/strings-in-block/class-html.php
brucepearson: @andreasciamanna I applied your suggestion with a small fix.

src/strings-in-block/class-html.php
```diff
@@ -174,4 +174,38 @@ private function get_dom_handler( \WP_Block_Parser_Block $block ) {
...
+		$translation = $this->getTranslation( $text, $lang, $block, $string_translations );
+		if ( $translation ) {
+			$block = self::update_string_in_innerContent( $block, $element, $translation );
+			$dom_handle->setElementValue( $element, $translation );
+		}
+
+		return $block;
+	}
+
+	private function getTranslation( $text, $lang, \WP_Block_Parser_Block $block, array $string_translations ) {
```
strategio, line 198: @brucepearson, don't you think we could also use this method for block attributes?

https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/blob/7d1dc300f95fdacefca3534c47093abf2662e527/src/strings-in-block/class-attributes.php#L161

I don't know if it's yet possible to have a shortcode inside a block attribute but I guess it will come soon :)

This `getTranslation` method could go in the parent class.

andreasciamanna (dismissed):

src/strings-in-block/class-html.php
brucepearson: @strategio I still have to go through all Toolset blocks. I'll add when/if it's needed.

strategio (dismissed):
+1 @brucepearson

dgwatkins (dismissed):
:+1:

PR #84, opened 2020-06-11
Fix edge case when there's only one block on the page and it contains…

brucepearson:
… a shortcode


strategio (approved):
+1 @brucepearson

gregWierzba (approved):
+1 @brucepearson

PR #85, opened 2020-07-07
Clean up unused shortcode strings when in Gutenberg blocks

brucepearson:
Clean up unused shortcode strings when in Gutenberg blocks - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmltm-3843

### DIFF
--- a/src/class-wpml-gutenberg-strings-registration.php
+++ b/src/class-wpml-gutenberg-strings-registration.php
@@ -48,25 +48,22 @@ class WPML_Gutenberg_Strings_Registration {
 	 */
 	public function register_strings( WP_Post $post, $package_data ) {
 		do_action( 'wpml_start_string_package_registration', $package_data );
+		do_action( 'wpml_start_GB_register_strings', $post, $package_data );
 
 		$this->leftover_strings = $original_strings = $this->string_translation->get_package_strings( $package_data );
 		$this->string_location  = 1;
 
-		$stringCleanUp = apply_filters( 'wpml_pb_get_string_clean_up', '', $post->ID );
-
 		$this->register_blocks(
 			WPML_Gutenberg_Integration::parse_blocks( $post->post_content ),
 			$package_data,
-			$post->ID,
-			$stringCleanUp
+			$post->ID
 		);
 
 		$current_strings = $this->string_translation->get_package_strings( $package_data );
 
 		$this->reuse_translations->find_and_reuse_translations( $original_strings, $current_strings, $this->leftover_strings );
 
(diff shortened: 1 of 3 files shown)


src/class-wpml-gutenberg-strings-registration.php
```diff
@@ -52,24 +52,29 @@ public function register_strings( WP_Post $post, $package_data ) {
 		$this->leftover_strings = $original_strings = $this->string_translation->get_package_strings( $package_data );
 		$this->string_location  = 1;
 
+		$stringCleanUp = apply_filters( 'wpml_pb_get_string_clean_up', '', $post->ID );
```
strategio: @brucepearson, please add a PHPDoc for this new filter hook. It's not obvious which return type we are expecting.

src/class-wpml-gutenberg-strings-registration.php
```diff
@@ -52,24 +52,29 @@ public function register_strings( WP_Post $post, $package_data ) {
...
-			$post->ID
+			$post->ID,
+			$stringCleanUp
 		);
 
 		$current_strings = $this->string_translation->get_package_strings( $package_data );
 
 		$this->reuse_translations->find_and_reuse_translations( $original_strings, $current_strings, $this->leftover_strings );
 
+		$stringCleanUp->cleanUp();
```
strategio: `$stringCleanUp` can be an empty string, so you should check it before to call the method.

strategio (changes requested):

andreasciamanna (dismissed):

src/class-wpml-gutenberg-strings-registration.php
brucepearson: @strategio I think I might change this and remove the filter and call a function directly. In reality $stringCleanUp will always be a StringCleanUp instance.

gordonsh24 (dismissed):

src/class-wpml-gutenberg-strings-registration.php
brucepearson: @strategio I refactored this to remove the dependency on `StringCleanUp`

src/class-wpml-gutenberg-strings-registration.php
brucepearson: @strategio I've refactored this.

andreasciamanna (dismissed):

src/class-wpml-gutenberg-strings-registration.php
strategio: Well done @brucepearson :)

PR #86, opened 2020-07-07
Try to get block translation from an external source also when the block does not have an xpath

strategio:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7341

### DIFF
--- a/src/strings-in-block/class-html.php
+++ b/src/strings-in-block/class-html.php
@@ -82,16 +82,13 @@ class HTML extends Base {
 			}
 			list( $block->innerHTML, ) = $dom_handle->getFullInnerHTML( $dom->documentElement );
 
-		} else {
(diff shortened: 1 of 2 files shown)


src/strings-in-block/class-html.php
```diff
@@ -82,16 +82,13 @@ public function update( \WP_Block_Parser_Block $block, array $string_translation
 			}
 			list( $block->innerHTML, ) = $dom_handle->getFullInnerHTML( $dom->documentElement );
 
-		} else {
+		} elseif ( isset( $block->blockName, $block->innerHTML ) && '' !== trim( $block->innerHTML ) ) {
```
strategio, line 85: This are the conditions we have inside `get_block_string_id`.

brucepearson (approved):
@strategio Looks good. +1

andreasciamanna (approved):

PR #87, opened 2020-07-20
Delegate also string registration if no string is found in a block

strategio:
It's not possible to check if a filter was called (at least not with `10up/wp_mock` on version `0.2.0` which is the last version compatible with PHP 5.6), I so don't have unit test for this change. But the whole issue will definitely require a CC test.

@brucepearson, please check this.

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7375


brucepearson (dismissed):
@strategio Looks good. +1

andreasciamanna (dismissed):

PR #89, opened 2020-08-13
Display HTML block translation field as a text area in CTE

mrudzki91:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7310

### DIFF
--- a/src/strings-in-block/class-html.php
+++ b/src/strings-in-block/class-html.php
@@ -167,14 +167,12 @@ class HTML extends Base {
 	 * @return ListBlock|StandardBlock|HtmlBlock
 	 */
 	private function get_dom_handler( \WP_Block_Parser_Block $block ) {
-		if ( self::LIST_BLOCK_NAME === $block->blockName ) {
-			return new ListBlock();
-		}
-		if ( self::HTML_BLOCK_NAME === $block->blockName ) {
-			return new HtmlBlock();
-		}
+		$class = wpml_collect( [
+			self::LIST_BLOCK_NAME => ListBlock::class,
+			self::HTML_BLOCK_NAME => HtmlBlock::class,
+		] )->get( $block->blockName, StandardBlock::class );
 
-		return new StandardBlock();
+		return new $class();
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


andreasciamanna (dismissed):

src/strings-in-block/dom-handler/html-block.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace WPML\PB\Gutenberg\StringsInBlock\DOMHandler;
+
+class HtmlBlock extends DOMHandle {
```
strategio: @mrudzki91, if you extend `HtmlBlock` from `StandardBlock`, you just have to write the method `getInnerHTML`.

src/strings-in-block/class-html.php
```diff
@@ -162,12 +164,15 @@ private function get_block_queries( \WP_Block_Parser_Block $block ) {
...
-	 * @return ListBlock|StandardBlock
+	 * @return ListBlock|StandardBlock|HtmlBlock
 	 */
 	private function get_dom_handler( \WP_Block_Parser_Block $block ) {
 		if ( self::LIST_BLOCK_NAME === $block->blockName ) {
 			return new ListBlock();
 		}
+		if ( self::HTML_BLOCK_NAME === $block->blockName ) {
+			return new HtmlBlock();
+		}
```
strategio: @mrudzki91, I suggest this smart pattern to replace kind of switch:

```php
$class = wpml_collect( [
    self::LIST_BLOCK_NAME => ListBlock::class,
    self::HTML_BLOCK_NAME => HtmlBlock::class,
] )->get( $block->blockName, StandardBlock::class );

return new $class();
```

Credits: @brucepearson :)

strategio (commented):
@mrudzki91, I have 2 small suggestions to reduce the quantity of code and duplication.

src/strings-in-block/dom-handler/html-block.php
mrudzki91: @strategio you right, it's done

src/strings-in-block/class-html.php
mrudzki91: @strategio great idea, thanks! :)

strategio (approved):
+1 nice job @mrudzki91

brucepearson (approved):

PR #91, opened 2021-02-05
Fix problem with html comments in gutenberg block content

mrudzki91:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmltm-4021

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -8,6 +8,10 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - uses: nanasess/setup-php@master
+        with:
+          php-version: '7.4'
+
       - name: Set github token for composer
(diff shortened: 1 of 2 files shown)


andreasciamanna (dismissed):

andreasciamanna:
@mrudzki91 Automated checks seem to be failing

andreasciamanna (approved):

dgwatkins (commented):
@mrudzki91 Isn't it the same to add another line to the strtr() array argument instead of adding str_replace()?

mrudzki91:
> @mrudzki91 Isn't it the same to add another line to the strtr() array argument instead of adding str_replace()?

@dgwatkins It isn't because the incorrect '--/>' string is produced by `strtr()` in some cases. That's why we need to run another replace on it's result.

dgwatkins (approved):
:+1: @mrudzki91

PR #93, opened 2021-02-25
Preserve newlines when applying translation - https://onthegosystems.…

brucepearson:
Preserve newlines when applying translation - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-7669

### DIFF
--- a/src/strings-in-block/class-html.php
+++ b/src/strings-in-block/class-html.php
@@ -2,7 +2,7 @@
 
 namespace WPML\PB\Gutenberg\StringsInBlock;
 
-use WPML\FP\Maybe;
+use WPML\FP\Str;
 use WPML\FP\Obj;
 use WPML\PB\Gutenberg\StringsInBlock\DOMHandler\HtmlBlock;
 use WPML\PB\Gutenberg\StringsInBlock\DOMHandler\StandardBlock;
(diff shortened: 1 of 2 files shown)


src/strings-in-block/class-html.php
```diff
@@ -211,12 +211,28 @@ private function getTranslation( $text, $lang, \WP_Block_Parser_Block $block, ar
...
+				return self::preserveNewLines( $text, $string_translations[ $string_id ][ $lang ]['value'] );
 			} else {
 				return null;
 			}
 		} else {
 			return $translationFromPageBuilder;
 		}
 	}
+
+	private static function preserveNewLines( $original, $translation ) {
```
brucepearson, line 223: This adds leading and trailing newlines to the translation if they are missing and in the original. It's required so the embeds work correctly. To me it seems safe to do it for all blocks and not just embeds.

@strategio What do you think?

strategio: I agree with you @brucepearson, I believe this can apply to all blocks.

src/strings-in-block/class-html.php
```diff
@@ -211,12 +211,28 @@ private function getTranslation( $text, $lang, \WP_Block_Parser_Block $block, ar
...
+		$endsWith = function ( $find, $s ) {
+			return Str::sub( - Str::len( $find ), $s ) === $find;
```
strategio, line 225: @brucepearson, I was looking at the `Str::sub` and `Str::len` and noticed that we are not using multi-bytes functions. This is something that should be adjusted in the FP library I think, or do you know the reason for not using it?

strategio (approved):
Looks nice @brucepearson +1

src/strings-in-block/class-html.php
brucepearson: @strategio I've added a ticket - https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-8099

dgwatkins (approved):
Looks good @brucepearson

PR #95, opened 2021-06-29
Handle gutenberg blocks in Widget Block

gordonsh24:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlst-2470

### DIFF
--- a/src/class-wpml-gutenberg-strings-registration.php
+++ b/src/class-wpml-gutenberg-strings-registration.php
@@ -125,6 +125,55 @@ class WPML_Gutenberg_Strings_Registration {
 		}
 	}
 
+	/**
+	 * @param array $blocks
+	 * @param array $package_data
+	 */
+	public function register_blocks_from_widget( array $blocks, array $package_data ) {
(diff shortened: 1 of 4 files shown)


gordonsh24:
@strategio can you look at this first?

src/class-wpml-gutenberg-strings-registration.php
```diff
@@ -125,6 +125,55 @@ private function register_blocks( array $blocks, array $package_data, $post_id )
 		}
 	}
 
+	/**
+	 * @param array $blocks
+	 * @param array $package_data
+	 */
+	public function register_blocks_from_widget( array $blocks, array $package_data ) {
```
strategio: Except the filters, it seems to be the same code as in `WPML_Gutenberg_Strings_Registration::register_blocks`. Did you try to call directly `WPML_Gutenberg_Strings_Registration::register_blocks` with `$post_id` equal to `0` or `null`?

strategio: @gordonsh24

PR #96, opened 2021-06-30
Add support for block widgets translation

strategio:
https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlst-2470

### DIFF
--- a/src/Widgets/Block/DisplayTranslation.php
+++ b/src/Widgets/Block/DisplayTranslation.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace WPML\PB\Gutenberg\Widgets\Block;
+
+use WPML\Element\API\Languages;
+use WPML\FP\Fns;
+use WPML\LIB\WP\Hooks;
+use function WPML\Container\make;
+use function WPML\FP\spreadArgs;
+
+class DisplayTranslation implements \IWPML_Frontend_Action, \WPML\PB\Gutenberg\Integration {
+
+	const PRIORITY_BEFORE_REMOVE_BLOCK_MARKUP = 0;
+
+	public function add_hooks() {
+		$getStringsFromMOFile = Fns::memorize( [ Strings::class, 'fromMo' ] );
+
+		Hooks::onFilter( 'widget_block_content', self::PRIORITY_BEFORE_REMOVE_BLOCK_MARKUP )
+		     ->then( spreadArgs( function ( $content ) use ( $getStringsFromMOFile ) {
+			     $strings = $getStringsFromMOFile( \get_locale() );
+
+			     return make( \WPML_Gutenberg_Integration::class )
+				     ->replace_strings_in_blocks( $content, $strings, Languages::getCurrentCode() );
+		     } ) );
+	}
+}
--- a/src/Widgets/Block/RegisterStrings.php
+++ b/src/Widgets/Block/RegisterStrings.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace WPML\PB\Gutenberg\Widgets\Block;
+
+use WPML\FP\Fns;
+use WPML\FP\Logic;
+use WPML\FP\Lst;
+use WPML\FP\Obj;
+use WPML\LIB\WP\Hooks;
+use function WPML\Container\make;
+use function WPML\FP\pipe;
+use function WPML\FP\spreadArgs;
+
+class RegisterStrings implements \IWPML_REST_Action, \WPML\PB\Gutenberg\Integration {
+	public function add_hooks() {
+		$registerStrings = Fns::memorize( function ( $oldValue, $newValue ) {
+			$gutenbergIntegration = make( \WPML_Gutenberg_Integration::class );
+			$blocks               = $this->getBlocks( $gutenbergIntegration, $newValue );
+
+			$gutenbergIntegration->register_strings_from_widget( $blocks, Strings::createPackage() );
+		} );
+
+		Hooks::onAction( 'update_option_widget_block', 10, 2 )
+		     ->then( spreadArgs( $registerStrings ) );
+	}
+
+	private function getBlocks( $gutenbergIntegration, $options ) {
+		$getContent = Logic::ifElse( 'is_scalar', Fns::always( null ), Obj::prop( 'content' ) );
+
+		return wpml_collect( $options )
+			->map( $getContent )
+			->filter()
+			->unique()
+			->map( [ $gutenbergIntegration, 'parse_blocks' ] )
+			->flatten( 1 )
+			->toArray();
+	}
+}
--- a/src/Widgets/Block/Strings.php
+++ b/src/Widgets/Block/Strings.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace WPML\PB\Gutenberg\Widgets\Block;
+
+use WPML\Element\API\Languages;
+use WPML\FP\Maybe;
+use WPML\FP\Obj;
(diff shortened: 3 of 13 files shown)


src/Widgets/Block/DisplayTranslation.php
```diff
@@ -0,0 +1,24 @@
...
+use WPML\LIB\WP\Hooks;
+use function WPML\Container\make;
+use function WPML\FP\spreadArgs;
+
+class DisplayTranslation implements \IWPML_Frontend_Action, \WPML\PB\Gutenberg\Integration {
+
+	const PRIORITY_BEFORE_REMOVE_BLOCK_MARKUP = 0;
+
+	public function add_hooks() {
+		Hooks::onFilter( 'widget_block_content', self::PRIORITY_BEFORE_REMOVE_BLOCK_MARKUP )
```
strategio: This is a new filter added in WP 5.8.

My suggestion is to get the widget translations directly from the MO file and use it to replace strings in blocks. It does not add any DB query, and the whole logic will be simpler.

It's also consistent with the way we are currently translating strings in legacy widgets (on-the-fly).

src/Widgets/Block/Strings.php
```diff
@@ -0,0 +1,60 @@
...
+
+	const DOMAIN = self::PACKAGE_KIND_SLUG . '-' . self::PACKAGE_NAME;
+
+	/**
+	 * @param string $locale
+	 *
+	 * @return array
+	 * @throws \WPML\Auryn\InjectionException
+	 */
+	public static function fromMo( $locale ) {
```
strategio, line 27: We could cache the result of this function.

src/class-wpml-gutenberg-integration-factory.php
```diff
@@ -1,29 +1,55 @@
...
-		/**
-		 * @var SitePress $sitepress
-		 * @var wpdb      $wpdb
-		 */
-		global $sitepress, $wpdb;
+		$integrations = new WPML\PB\Gutenberg\Integration_Composite();
 
-		$config_option = new WPML_Gutenberg_Config_Option();
+		$mainIntegration = $this->create_gutenberg_integration();
+		share( [ $mainIntegration ] );
```
strategio, line 13: By sharing this instance of the main integration class, we can reuse it in the other classes by just `make( \WPML_Gutenberg_Integration::class );`.

strategio:
@gordonsh24, I started by porting the code from ST to this package. At this point I did some tests and it was not working properly for me. The option was translated in the DB, but I could see some issues: The instance indexes where not matching, and even if I tried to change it manually, it was failing to show on the translated frontend page. Moreover, sometimes I was just loosing the original instance, not sure exactly how, maybe when you switch the admin language on the Widgets admin page.

So I decided to try to "translate on the fly" strategy (starting on commit 532a202) and it works pretty well with no extra DB query. It relies on the custom MO files created for the widgets package.

I let you review all my changes and improve it (it requires a second pass). I just adjusted the existing tests but it also requires more coverage, I'll let you check.

I also create a counter part in ST (https://git.onthegosystems.com/wpml/wpml-string-translation/-/merge_requests/1294) which only updates the dependencies (but tests are failing there, not sure why).

It's in your hands now!

gordonsh24:
@strategio It looks great. Thanks a lot for handling it. I just added two missing tests and made a few small adjustments.

gordonsh24:
@BruceOnTheGo @andreasciamanna can you review it?

src/Widgets/Block/RegisterStrings.php
```diff
@@ -0,0 +1,40 @@
...
+
+		$fn = pipe(
+			Fns::map( $getContent ),
+			Fns::filter( Fns::identity() ),
+			Fns::unary( 'array_unique' ),
+			Fns::map( [ $gutenbergIntegration, 'parse_blocks' ] ),
+			Lst::flattenToDepth( 1 )
+		);
+
+		return $fn( $options );
```
brucepearson: @strategio I would probably use a collection here:

```php
$getContent = Logic::ifElse( 'is_scalar', Fns::always( null ), Obj::prop( 'content' ) );

return wpml_collect( $options )
   ->map( $getContent )
   ->filter()
   ->unique()
   ->map( [ $gutenbergIntegration, 'parse_blocks' ] )
   ->flatten(1)
   ->toArray();

```

src/Widgets/Block/Strings.php
```diff
@@ -0,0 +1,64 @@
...
+		$moFilePath = Manager::getSubdir() . '/' . $domain . "-$locale.mo";
+
+		if ( ! file_exists( $moFilePath ) ) {
+			return [];
+		}
+
+		$mo = make( \MO::class );
+		$mo->import_from_file( $moFilePath );
+
+		return $mo->entries;
```
brucepearson: @strategio You could do:
```php
return Maybe::of( Manager::getSubdir() . '/' . $domain . "-$locale.mo" )
             ->filter( 'file_exists' )
             ->map( [ make( \MO::class ), 'import_from_file' ] )
             ->map( Obj::prop( 'entities' ) )
             ->getOrElse( [] );
```

src/Widgets/Block/Strings.php
```diff
@@ -0,0 +1,64 @@
...
+			->map( function( $value ) use ( $langCode ) {
+				return [
+						$langCode => [
+							'value'  => $value,
+							'status' => ICL_STRING_TRANSLATION_COMPLETE,
+						],
+					];
+				} )
+			->toArray();
+	}
```
brucepearson, line 44: @strategio You could do:
```php
$encode = curryN( 2, function( $langCode, $value ) {
				return [
						$langCode => [
							'value'  => $value,
							'status' => ICL_STRING_TRANSLATION_COMPLETE,
						],
					];
				});

return wpml_collect( self::loadStringsFromMOFile(self::DOMAIN, $locale) )
			->map( Obj::path( [ 'translations', 0 ] ) )
			->filter()
			->map( $encode( $langCode ) )
			->toArray();
```php

File: src/class-wpml-gutenberg-integration.php [hunk 7]
```diff
@@ -100,18 +104,30 @@ public function string_translated(
...
+	 *
+	 * @return string
+	 */
+	public function replace_strings_in_blocks( $content, $string_translations, $lang ) {
+		$blocks = self::parse_blocks( $content );
 
-			wpml_update_escaped_post( [ 'ID' => $translated_post_id, 'post_content' => $content ], $lang );
+		$blocks = $this->update_block_translations( $blocks, $string_translations, $lang );
+
+		$newContent = '';
```
brucepearson: @strategio Maybe:
```php
return Fns::reduce( Str::concat(), '', Fns::map( [ $this, 'render_block ], $blocks ) );
```

brucepearson (dismissed):
@strategio @gordonsh24 Only some minor comments that are not vital. Nice work!

andreasciamanna (dismissed):

src/Widgets/Block/Strings.php
gordonsh24: @brucepearson yes it looks better in this version. Changed.

src/Widgets/Block/Strings.php
gordonsh24: Done

src/class-wpml-gutenberg-integration.php
gordonsh24: @brucepearson nice! Done.

gregWierzba (approved):
+1

andreasciamanna (approved):

PR #100, opened 2021-07-06
Add support for media translation of block widgets

strategio:
The tests are failing because it requires a small adjustment in `makeMock` (https://git.onthegosystems.com/wpml-packages/core-api/-/merge_requests/77).

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-8670

### DIFF
--- a/src/Widgets/Block/DisplayTranslation.php
+++ b/src/Widgets/Block/DisplayTranslation.php
@@ -19,6 +19,12 @@ class DisplayTranslation implements \IWPML_Frontend_Action, \WPML\PB\Gutenberg\I
 		     ->then( spreadArgs( function ( $content ) use ( $getStringsFromMOFile ) {
 			     $strings = $getStringsFromMOFile( \get_locale() );
 
+			     $imageTranslator = make( \WPML_Media_Translated_Images_Update::class );
+
+			     if ( $imageTranslator ) {
+				     $content = $imageTranslator->replace_images_with_translations( $content, Languages::getCurrentCode() );
+			     }
+
 			     return make( \WPML_Gutenberg_Integration::class )
 				     ->replace_strings_in_blocks( $content, $strings, Languages::getCurrentCode() );
 		     } ) );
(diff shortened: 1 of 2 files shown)


strategio:
@OnTheGoSystems/wpml

src/Widgets/Block/DisplayTranslation.php
```diff
@@ -19,6 +19,12 @@ public function add_hooks() {
 		     ->then( spreadArgs( function ( $content ) use ( $getStringsFromMOFile ) {
 			     $strings = $getStringsFromMOFile( \get_locale() );
 
+			     $imageTranslator = make( \WPML_Media_Translated_Images_Update::class );
```
brucepearson, line 22: @strategio I think this should be moved to Media plugin and Media plugin should hook `widget_block_content` before this.

gordonsh24: @brucepearson We discussed it with Pierre before when we were working on the first Gutenberg task. I also initially put my code into ST, but Pierre explained that it is better to keep it here. 

1. Now, the whole Gutenberg related code is stored in one place
2. if something changes just in Gutenberg, the compatibility is able to release a fix without WPML

gordonsh24 (approved):

src/Widgets/Block/DisplayTranslation.php
brucepearson: ok

brucepearson (approved):

src/Widgets/Block/DisplayTranslation.php
strategio: I have been hesitating, both ways have pros and cons...

In the end, I think it's more correct to move the media filtering to WPML Media.

strategio:
Closing here, I moved the code to https://git.onthegosystems.com/wpml/wpml-media-translation/-/merge_requests/350.

PR #101, opened 2021-07-08
Improve core/search block to search in a current lang when lang as pa…

gordonsh24:
…rameter is set

https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-8668

### DIFF
--- a/src/Widgets/Block/Search.php
+++ b/src/Widgets/Block/Search.php
@@ -1,0 +1,19 @@
+<?php
+
+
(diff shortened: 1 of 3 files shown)


gordonsh24:
We need to apply the same logic as we do for a legacy search form via `add_filter( 'get_search_form', array( $this, 'get_search_form_filter' ) );`

nemanjatan (approved):
Nice!

andreasciamanna (approved):

PR #102, opened 2021-08-04
Fix/handle multitag wrap

yanmorinokamca:
Fix problem with inner html like: `<a><b><c></c></b></a>`.  Since `></`  is the first match in the center, as `></`, the explode() will create many parts for b and a. This patch limit to 2 parts, to make `<a><b><c>` and `</c></b></a>`. 

In render_inner_HTML_with_guess_parts, it always assume that the array is of length 2 (doesn't use anything else).

(edit from Pierre: https://onthegosystems.myjetbrains.com/youtrack/issue/wpmlcore-8769)

### DIFF
--- a/src/class-wpml-gutenberg-integration.php
+++ b/src/class-wpml-gutenberg-integration.php
@@ -292,12 +292,15 @@ class WPML_Gutenberg_Integration implements \WPML\PB\Gutenberg\Integration {
 
 				if ( count( $matches ) === 1 ) {
 					$parts = explode( $matches[0], $inner_HTML );
-					if ( count( $parts ) === 2 ) {
+					if ( count( $parts ) >= 2 ) {
+						$partsRight = array_slice( $parts, 1 );
 						$match_mid_point = 1 + ( mb_strlen( $matches[0] ) - 3 ) / 2;
 						// This is the first ">" char plus half the remaining between the tags
 
-						$parts[0] .= mb_substr( $matches[0], 0, $match_mid_point );
-						$parts[1] = mb_substr( $matches[0], $match_mid_point ) . $parts[1];
+						$parts = array(
+							$parts[0] . mb_substr( $matches[0], 0, $match_mid_point ),
+							mb_substr( $matches[0], $match_mid_point ) . implode( '></', $partsRight ),
+						);
 					}
 				}
 				break;
--- a/src/class-wpml-gutenberg-integration.php
+++ b/src/class-wpml-gutenberg-integration.php
@@ -291,16 +291,13 @@ class WPML_Gutenberg_Integration implements \WPML\PB\Gutenberg\Integration {
 				preg_match( '#>\s*</#', $inner_HTML, $matches );
 
 				if ( count( $matches ) === 1 ) {
-					$parts = explode( $matches[0], $inner_HTML );
-					if ( count( $parts ) >= 2 ) {
-						$partsRight = array_slice( $parts, 1 );
+					$parts = explode( $matches[0], $inner_HTML, 2 );
+					if ( count( $parts ) === 2 ) {
 						$match_mid_point = 1 + ( mb_strlen( $matches[0] ) - 3 ) / 2;
 						// This is the first ">" char plus half the remaining between the tags
 
-						$parts = array(
-							$parts[0] . mb_substr( $matches[0], 0, $match_mid_point ),
-							mb_substr( $matches[0], $match_mid_point ) . implode( $matches[0], $partsRight ),
-						);
+						$parts[0] .= mb_substr( $matches[0], 0, $match_mid_point );
+						$parts[1] = mb_substr( $matches[0], $match_mid_point ) . $parts[1];
 					}
 				}
 				break;
--- a/src/class-wpml-gutenberg-integration.php
+++ b/src/class-wpml-gutenberg-integration.php
@@ -299,7 +299,7 @@ class WPML_Gutenberg_Integration implements \WPML\PB\Gutenberg\Integration {
 
 						$parts = array(
 							$parts[0] . mb_substr( $matches[0], 0, $match_mid_point ),
-							mb_substr( $matches[0], $match_mid_point ) . implode( '></', $partsRight ),
+							mb_substr( $matches[0], $match_mid_point ) . implode( $matches[0], $partsRight ),
 						);
 					}
 				}
--- a/tests/phpunit/tests/test-wpml-gutenberg-integration.php
+++ b/tests/phpunit/tests/test-wpml-gutenberg-integration.php
@@ -454,7 +454,8 @@ class Test_WPML_Gutenberg_Integration extends OTGS_TestCase {
 				"<div class=\"wp-block-column\">\r\n\r\n</div>",
 				"<div class=\"wp-block-column\">\r\n",
 				"\r\n</div>"
-			)
+			),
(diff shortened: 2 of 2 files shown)


strategio:
@yanmorinokamca, I think we can strengthen our unit tests with your suggested fix.

Could you please extend the data provider in [\Test_WPML_Gutenberg_Integration::inner_html_provider](https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/blob/master/tests/phpunit/tests/test-wpml-gutenberg-integration.php#L448) by adding one more line:

```php
array( '<div class="wp-block-column"><b><span></span></b></div>', '<div class="wp-block-column"><b><span>', '</span></b></div>' ),
```

Do you have a real-life scenario to reproduce this issue (could you share some steps to follow on a site)?

Thanks for your contribution!

yanmorinokamca:
We have a website with custom gutenberg blocks, and use translation from french to english and spanish. The first block has two custom properties / attributes, like `<okam-tabs tabs-array={someJSONArray} tabs-aria-label={string}> <InnerBlocks.Content /> </okam-tabs>`. It add the property keys inside the `<!--wp -->` comment and inside the attributes. It was not saving correctly, so I've added a key/xpath inside a wpml-config.xml. 

I've also added this patch because the Collection in vendor/wpml/page-builders-gutenberg/src/strings-in-block/class-collection.php was updating the block->innerHTML but not the innerContent, and later inside render_block, WPML_Gutenberg_Integration::render_inner_HTML will only use the innerContent if $block->innerContent is set instead of using the new innerHTML with render_inner_HTML_with_guess_parts.

```
add_filter('wpml_update_strings_in_block', function ($block) {
    if (isset($block->attrs['translatedWithWPMLTM']) && $block->attrs['translatedWithWPMLTM'] === '1') {
        $block->innerContent = null;
    }   
    return $block;
}, 10, 1); 
```

So that's for the first block, after this patch, I found the second error with a custom block that wrap an url as `<div><a href={href}><div> <InnerBlocks.Content />  </div></a></div>`
In this case, I've added a gutenber-block in wpml-config.xml on the xpath //div/a/@href.
And this is why I found the bug in this fix.

strategio:
[Since WP 5.0 (or Gutenberg 4.2)](https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/commit/3af51a7217fa544c20100cb554c46925fe4fcca5#diff-30e38340deb4c7fca57bab6441d8780b4f06d3bb4f6112037112fc45bf633104R216), this branch of the code should not be used anymore, that's why I was wondering.

Now I see that you are explicitly unsetting `$block->innerContent` with your custom filter, so this explains why.

My suggestion is to move this fix to WPML 4.5.0 (on the `develop` branch) so that we'll might have a better beta testing (to avoid breaking things).

@yanmorinokamca could you please re-build your branch to target `develop`? Or I can do this if you prefer...

yanmorinokamca:
It seems like they are using Gutenberg 10.6.2, WPML 4.4.10, Wordpress 5.7.2 on that website, not sure why I fall in this code when debugging with xdebug. 
I'm not sure how to re-build the branch target, so I prefer if you do it, I've cloned the repository, and there was only master in the git branch listing, so I forgot to switch to develop. Oops... Should have done a `git branch -a`.

strategio:
>It seems like they are using Gutenberg 10.6.2, WPML 4.4.10, Wordpress 5.7.2 on that website, not sure why I fall in this code when debugging with xdebug.

It's because the property `$block->innerContent` was added in Gutenberg 4.2 (or WP 5.0) and we check it to decide whether to use the old or the new method. But you added your custom filter `wpml_update_strings_in_block` to reset this property. I believe this is the reason why.

It's not clear to me why you are removing this property. Maybe what we are trying to fix here is a side effect of your filter... Could you re-check again?

To be honest, it might be wiser to start from a [support ticket on wpml.org](https://wpml.org/forums/forum/english-support/) where you can share a copy of your site. That way we can investigate on a real-life scenario and make sure the fix is not just a hack.

Even though your fix looks legit, we try to have a clear picture of the problems we fix, so we don't swap a bug for another.

If you agree with this, please start a forum thread and share the link here. Then I will continue the conversation with you on the forum :)

Thanks @yanmorinokamca!

yanmorinokamca:
Hi, Concerning the first bug (which force me to set innerContent to null in a filer), I've retrace the code with Wordpress 5.8, Gutenberg 11.2.1 and WPML 4.4.11 and I'm still getting the bug without this filter.

If there is block queries (xpath attribute to change) for a block, and the block contains innerBlocks, 
* WPML\PB\Gutenberg\StringsInBlock\HTML will update the $block->innerHTML  but not $block->innerContent. 
* Then, it returns to WPML\PB\Gutenberg\StringsInBlock\Collection\update where the innerHTML is different from originalInnerHTML (we got the translatedWithWPMLTM = '1' there)
* Then back to WPML_Gutenberg_Integration::update_block_translations
* Back to WPML_Gutenberg_Integration::replace_stirings_in_blocks
* Which call WPML_Gutenberg_Integration::render_block on each blocks
* Which call WPML_Gutenberg_Integration::render_inner_HTML
* Which render the innerContent in priority if there is innerBlocks. The innerContent was not updated by the xpath logic, therefore, we don't get the new updated attributes in the content.

I will open a ticket on wpml with the wpml-config.xml files and an example of the content generated in wp_posts.post_content

yanmorinokamca:
Well, well, I've searched and searched and got why the innerContent is not replaced. In WPML\PB\Gutenberg\StringsInBlock\HTML::update_string_in_innerContent, the xpath try to replace a DOMAttr, so the search value is a preg_quote of the $element->nodeValue : `"\["INNOVATION", "DÉVELOPPEMENT DURABLE", "CONFORMITÉ", "RÉSEAU MONDIAL9"\]"`   
Or, the $block->innerContent is `"
<okam-tabs class="wp-block-okam-okam-tabs" tabs-array="[&quot;INNOVATION&quot;, &quot;DÉVELOPPEMENT DURABLE&quot;, &quot;CONFORMITÉ&quot;, &quot;RÉSEAU MONDIAL9&quot;]" tabs-aria-label="Go to tab ">"`.

Since the attribute $element->nodeValue is not using htmlspecialchars() to replace " into `&quot;`  the preg_replace can't change the inner_content with the &quot;. The innerHTML is correctly replace but not the innerContent.

I've tried that patch for DOMAttr only, I will create a new branch (from develop) with this fix later:
```
		if ( $element instanceof \DOMAttr ) {
			$search_value = preg_quote( htmlspecialchars($element->nodeValue), '/' );
			$search = '/(")(' . $search_value . ')(")/';
			$translation = htmlspecialchars($translation);
		} else {
			$search_value = preg_quote( $element->nodeValue, '/' );
			$search = '/(>)(' . $search_value . ')(<)/';
		}
```

I'm using htmlspecialchars() here, but maybe you will prefer to use another function, or use the good encoding parameters.

strategio:
@yanmorinokamca, is this PR still needed or https://github.com/OnTheGoSystems/wpml-page-builders-gutenberg/pull/103 is enough?

yanmorinokamca:
I think that the patch still fix a bug in this function, even if my setup don't call it now.

strategio:
Well, I prefer to not touch this code now if we don't have a concrete reason.

I will close this PR, thanks for your efforts @yanmorinokamca!

